PR #32, opened 2015-01-16
fix #31: handle discogs tracks in case of sub-tracks

### LINKED ISSUE #31: Discogs: handle tracks in case of sub-tracks
In case of sub-tracks (eg: http://api.discogs.com/releases/5880212), current code ignores discogs tracks.

I will propose a pull request which takes the first sub-track "position" and makes things work nicely
(except that sub-tracks are not imported to musicbrainz, but musicbrainz does not handle it?)

cu

pixel:
in case of sub-tracks (at least for http://api.discogs.com/releases/5880212 ),
json field "position" is "" so previous code ignored the track.

The solution here is to use first sub-track position which is good enough to detect change of CD numbers.

### DIFF
--- a/discogs_importer.user.js
+++ b/discogs_importer.user.js
@@ -371,6 +371,10 @@ function parseDiscogsRelease(data) {
         // Track position and release number
         var trackPosition = discogsTrack.position;
 
+        if (trackPosition == "" && discogsTrack.sub_tracks) {
+            trackPosition = discogsTrack.sub_tracks[0].position;
+        }
+
         // Skip special tracks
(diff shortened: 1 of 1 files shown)


murdos:
At least with the example you provided, the result is wrong.
From my understanding of the [back cover](http://www.discogs.com/release/5880212) the first CD should have 12 tracks, not 4. 
So each subtrack should be considered as a track for MusicBrainz import, probably by concataining titles of index track and subtrack, and index tracks should be skipped.

pixel:
After double checking on the CD, you're plain right /o\

Issue #6, opened 2013-07-15
Possibility to add the disc titles and release date for albums


Rouzax:
I also posted this on http://userscripts.org/topics/126546 but I'm not sure that is monitored.
First of all many thanks for this wonderful script it saves heaps of time and frustration.

Would it be possible to add the disc titles and release date for albums?

For instance this release, http://www.discogs.com/release/3979680 has a title for the different CD's 
CD1 = Bar: Mixed By Nathan Cozzetto 
CD2 = Club: Mixed By Dan Van 
CD3 = After Party: Mixed By Carl Hanaghan 
You have to copy those by yourself 
http://musicbrainz.org/release/628d2a34-6e7d-4d...

Also the Release event Date is not populated, would that be possible?

zas:
Disc titles are now supported, and release date was since a long time. Issue fixed.

PR #44, opened 2015-06-03
Discogs: import disc titles when possible

zas:
https://github.com/murdos/musicbrainz-userscripts/issues/6

Among releases tested:
http://www.discogs.com/release/7079425
http://www.discogs.com/release/7084165
http://www.discogs.com/release/3979680


murdos:
Looks fine, thanks!

PR #45, opened 2015-06-03
Handle sub-tracks and their titles

zas:
https://github.com/murdos/musicbrainz-userscripts/issues/31

Tested on http://www.discogs.com/release/5880212

### DIFF
--- a/discogs_importer.user.js
+++ b/discogs_importer.user.js
@@ -357,7 +357,7 @@ function parseDiscogsRelease(data) {
         if (discogsTrack.type_ == 'heading') {
           heading = discogsTrack.title;
           return;
-        } else if (discogsTrack.type_ != 'track') {
+        } else if (discogsTrack.type_ != 'track' && discogsTrack.type_ != 'index') {
           return;
         }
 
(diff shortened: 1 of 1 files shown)


murdos:
Thanks.

Just for the record, http://www.discogs.com/release/5880212 is a wrong example, since this release should use heading tracks instead.
http://www.discogs.com/release/1156598 is a better example.

discogs_importer.user.js
```diff
@@ -386,6 +386,29 @@ function parseDiscogsRelease(data) {
         // Track position and release number
         var trackPosition = discogsTrack.position;
 
+        // Handle sub-tracks
+        if (trackPosition == "" && discogsTrack.sub_tracks) {
+            trackPosition = discogsTrack.sub_tracks[0].position;
```
murdos: Use first subtrack position is not perfect, but better than nothing.

murdos:
Another improvement could be to compute the track length by summing length of subtracks.

zas:
> Another improvement could be to compute the track length by summing length of subtracks.
http://www.discogs.com/release/5880212 has duration provided in 'index' section, but yes, http://www.discogs.com/release/1156598 may need this feature ;)

discogs_importer.user.js
zas: Yes, this is a bit hacky, but discogs data format doesn't help ;)

zas: Note: track position isn't correct in all cases,  ie. http://www.discogs.com/release/1156598 will have track B in release editor set as track 5. I found various weird formats in positions like 'AAa' or 'Ab' or 'AA1'... the current script doesn't handle those.

PR #49, opened 2015-06-08
Mblinks optimization

murdos:
Mechanism to handle duplicated requests to MusicBrainz webservice: if a link is displayed in many places, only one Ajax request will be made, and it will run all registered handlers that have been merged in an array.

### DIFF
--- a/lib/mblinks.js
+++ b/lib/mblinks.js
@@ -25,16 +25,17 @@ var MBLinks = function (cachekey, expiration) {
   }();
 
   this.ajax_requests = {
-    // properties: "key": {handlers: [], next: property}
+    // properties: "key": {handler: function, next: property, context: {}}
     first: "",
     last: "",
     empty: function() {return this.first == "";},
-    push: function(key, handler) {
+    push: function(key, handler, context) {
         if (key in this) {
-            this[key].handlers.push(handler);
+            this[key].handler = handler;
+            this[key].context = context;
         }
         else {
-            this[key] = {handlers: [handler], next: ""};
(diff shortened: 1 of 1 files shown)


lib/mblinks.js
```diff
@@ -100,7 +130,13 @@ var MBLinks = function (cachekey, expiration) {
...
       });
     } else {
-      mblinks.ajax_requests.push($.proxy(function () {
+      // Merge with previous context if there's already a pending ajax request
+      var handlers = [];
+      if (url in mblinks.ajax_requests) {
+        handlers = mblinks.ajax_requests[url].context.handlers;
+      }
+      handlers.push(insert_func);
+      mblinks.ajax_requests.push(url, function () {
```
zas: I wonder if `url` is the correct thing to use here, i would use `query` defined as : 
```javascript
var query = mblinks.mb_server + '/ws/2/url?resource=' + encodeURIComponent(url) + '&inc=' + mb_type + '-rels';
```
We may do ajax requests for the same `url` but with different mb_type (it will be the case if we want to support MB places; which are labels on discogs side).

murdos: I agree for the proposed key, that makes sense.
Sorry for not responding sooner, I don't have much free time these days...
Feel free to make the modification if you want.

zas: No problem about the lack of time ;)
I changed to full query in https://github.com/murdos/musicbrainz-userscripts/commit/7c67333f92ea9b720b60e17f2de1aadbea07e75e

PR #52, opened 2015-06-12
BeatPort Pro: make it work again

zas:
Fixed:
- invisible import button
- more tracks than in the release
- release artist credit were the ones of the last track
- remove 'Original Mix' from track titles
- update jquery to recent version
- prevent jquery conflicts
- use relative paths for lib


CatQuest:
:thumbsup:

PR #55, opened 2015-06-13
bandcamp_importer: support bandcamp pages having their own domain

zas:
Currently, one has to add the domain to userscript preferences for pages provided by bandcamp
but under another domain (ie. http://bearthemammoth.com/album/yamadori)

This change makes the `@include` wider (it may match totally unrelated pages) and test for
a bandcamp-specific object before doing anything.

Please review and test, i'm still not sure if it is a good idea, but it simplifies things since nowaday a lot of artists/labels are using bandcamp engine with their own domains.

Another approach would be to make 2 userscripts with different include rules, sharing the same code.


murdos:
This seems fine, since you're testing the presence of unsafeWindow.TralbumData.

PR #56, opened 2015-06-14
Discogs importer: display links for direct search of entities on MB

zas:
While an entity is not found, display a small link which allows one to query MB for this entity.
For the case the discogs element may match more than one MB entities (ie. label or place), one search link for each is displayed (at the moment only in release > companies).

capture du 2015-06-14 22 01 56

### DIFF
--- a/discogs_importer.user.js
+++ b/discogs_importer.user.js
@@ -42,6 +42,7 @@ var mblinks = new MBLinks('DISCOGS_MBLINKS_CACHE', 7*24*60, '1'); // force refre
 $(document).ready(function(){
 
     MBImportStyle();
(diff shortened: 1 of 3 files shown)


murdos:
I've only been using this version of the script for a few days, and it seems OK.
(I've not really reviewed the code, I'm a bit lost with all the changes you made recently on the discogs script).

zas:
> I've only been using this version of the script for a few days, and it seems OK.

Thanks for testing it.

> (I've not really reviewed the code, I'm a bit lost with all the changes you made recently on the discogs script).
Well, yes ;)
You'll prolly have some clean up code sessions this summer.

Issue #57, opened 2015-06-15
Idea: Last.fm import script


liftarn:
Many independent artists (especially outside Europe and USA) use Last.fm to release their music so it would be nice to have a script to import them.
Last.fm uses release info from MusicBrainz so perhaps some kind of check to avoid roundtrips may be needed.

zas:
Do last.fm has "releases" by itself ? all i can find are redirected to third-party shops.

pankkake:
example: http://www.last.fm/music/Government+Alpha/FSCD12+-+GOVERNMENT+ALPHA+-+Groove+Test+Frequency+(2000)

Not to be confused to all the "releases" populated by crappy tags!

liftarn:
Compare for instance http://www.last.fm/music/LukHash/Digital+Memories that is a Last.fm release with http://www.last.fm/music/Birmingham+6/Mindhallucination that is imported from MusicBrainz. Notice that releases created by artists or labels have a release label while the imported don't.

zas:
https://github.com/murdos/musicbrainz-userscripts/pull/58

zas:
Merged ;)

PR #58, opened 2015-06-15
Add last.fm importer

zas:
- it doesn't import much data, since last.fm has almost none but the tracklist (a bit more if last.fm uses MB data, but in this case no point in importing...)
- it is very unclear which release last.fm is listing, editor have to take care of formats, release dates, label, etc...
- artist credits are somehow random, don't expect miracles

I don't think last.fm is a good source for releases, it should only be used in last resort.

Suggested by https://github.com/murdos/musicbrainz-userscripts/issues/57

### DIFF
--- a/lastfm_importer.user.js
+++ b/lastfm_importer.user.js
@@ -1,0 +1,111 @@
+﻿// ==UserScript==
+// @name           Import Last.fm releases to MusicBrainz
+// @namespace      https://github.com/murdos/musicbrainz-userscripts/
+// @description    One-click importing of releases from last.fm into MusicBrainz. PLEASE import releases from more reliable sources if possible.
+// @version        2015.06.15.0
+// @downloadURL    https://raw.githubusercontent.com/murdos/musicbrainz-userscripts/master/lastfm_importer.user.js
+// @updateURL      https://raw.githubusercontent.com/murdos/musicbrainz-userscripts/master/lastfm_importer.user.js
+// @include        /^https?:\/\/(?:www\.)?(?:last\.fm|lastfm\.(?:com\.br|com\.tr|at|com|de|es|fr|it|jp|pl|pt|ru|se))\/music\/[^\/]+/[^+][^\/]+/
+// @require        https://ajax.googleapis.com/ajax/libs/jquery/2.1.4/jquery.min.js
+// @require        lib/import_functions.js
+// @require        lib/logger.js
+// @require        lib/mbimportstyle.js
+// ==/UserScript==
+
+// prevent JQuery conflicts, see http://wiki.greasespot.net/@grant
+this.$ = this.jQuery = jQuery.noConflict(true);
+
+if (!unsafeWindow) unsafeWindow = window;
+
+$(document).ready(function(){
+  MBImportStyle();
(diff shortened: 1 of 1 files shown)


murdos:
I don't think scrapping the current page is the way to go.

I'd rather use the [last.fm API](http://www.lastfm.fr/api/show/album.getInfo) that furthermore provides MBIDs. For the 3 provided examples in #57 the release was already in MusicBrainz.

zas:
> I'd rather use the last.fm API that furthermore provides MBIDs. For the 3 provided examples in #57 the release was already in MusicBrainz.

Oh, i didnt even think about it ... I'm not familiar at all with last.fm
What about API keys and API Terms of Service/licenses ? If we use this API, the script user will have to have its own API key and therefore entering it somewhere ?

murdos:
Well, I'm wondering if there's a real need for an import script for lastfm.
As you said in your original comment, few data is available. And I've still not seen an example of a release available in lastfm but not in MB.

liftarn:
If there are few releases on Last.fm that already aren't in MusicBrainz it is only because me and others have worked hard on importing them manually. A script would make it so much easier.
Some examples not yet imported:
http://www.last.fm/music/Greg+Gibbs/The+Lights
http://www.last.fm/music/Greg+Gibbs/Infinite+Lute+Presents+Greg+Gibbs
http://www.last.fm/music/Greg+Gibbs/Raincoat
http://www.last.fm/music/Essence/Essence
http://www.last.fm/music/Greg+Gibbs/Greg+Gibbs+Song+a+Week
http://www.last.fm/music/Stromble+Fix/The+Early+Years
http://www.last.fm/music/Stromble+Fix/Toilsome+Days
http://www.last.fm/music/The+Silver+Wizard+Project/Live+@+the+Dry+Bar
http://www.last.fm/music/Saelynh/Aeon
http://www.last.fm/music/Superlauncher/Prism
http://www.last.fm/music/Superlauncher/Music+for+Films
http://www.last.fm/music/Itrium/The+Beast+Has+Risen+%28Demo%29
http://www.last.fm/music/Itrium/Within+Your+Hands+%28Demo%29
http://www.last.fm/music/Woody%27s+a+Girl/Under+Cover
http://www.last.fm/music/Woody%27s+a+Girl/Echoes
http://www.last.fm/music/Woody%27s+a+Girl/Big+Mac+Spacky
http://www.last.fm/music/Woody%27s+a+Girl/New
and so on...

murdos:
@liftarn: Thanks!

So the question is should we the API or not? Using it would allow identifying automatically artist for example with http://www.last.fm/music/Superlauncher/Prism but I don't know how to handle the API key issue.

Let's merge the proposed version for now, at least there's something usable ;-)

liftarn:
@murdos Actually Last.fm is still not able to tell artists with the same name apart (it's in progress I'm told) so that is not a reason. It could be used to tell if the release (or rather a release with the same name for artist and release) already is in MB (I have that info on my moderator panel).

liftarn:
It would be nice if the script could parse label and release date when available. Example: http://www.last.fm/music/For+All+Eternity/Solid+Ground

PR #68, opened 2015-06-28
Try to detect language of the release [experimental]

zas:
**For testing only, not yet finalized.**
If it works well enough, it may be moved to mbimport lib and used for all importers.

It uses a simple js code, and imho it should be modified to get the script too (the js lang detector is detecting the script too).
For all european languages we can easily set script to latin.


murdos:
I've been testing it for a few days (mainly for French releases), it seems to works fine if there's enough text. For singles with only 2 tracks the result is sometimes wrong.
I've also not tested release with mixed content that would normally end with [Multiple Langages].

zas:
> I've also not tested release with mixed content that would normally end with [Multiple Langages].

The threshold is quite high, i guess that mixed languages will end as unknown (in the lib), and therefore no language will be set for the release.

Freso:
> For all european languages we can easily set script to latin.

I thought Greece was using Greek and Russian and Ukrainian Cyrillic?

zas:
> I thought Greek was using Greek and Russian and Ukrainian Cyrillic?

Of course, excluding those.

Freso:
It might be good to rename the commit to denote that it's for the Discogs plugin and doesn't apply universally to all plugins. (Also, needs conflict resolving.)

PR #91, opened 2016-01-07
Issue 89: ignore repeated '.' or '-' in (buggy) track position

zas:
http://www.discogs.com/release/1520366 has a track with buggy position,
the script is placing the track at the wrong position because of this.

2-.23 	Le Carnaval 	0:26

2-.23 is now parsed as it was 2-23

This patch shouldn't affect non-buggy track positions.


murdos:
The change sounds safe, feel free to merge it.

PR #96, opened 2016-02-17
Move the MBz import box up between the Discogs Collection and Marketplace boxes

VxJasonxV:
This change shuffles the MBz box priorities up so that the 'Import into' and 'Search in MBz' buttons show up immediately under the Release information when present, above the Marketplace boxes.

On smaller windows, this box requires scrolling down to find, and is doubly obnoxious when having to come back to reference information and scroll back up.

Additionally, it also bumps the download and update URLs to use the raw.githubusercontent.com domain.

Tested by myself on Chrome 48.0.2564.103 (64-bit), using Tampermonkey 3.12.58, on OS X 10.11.2. Also mentioned in #musicbrainz awaiting further feedback.


murdos:
Merged, thanks!

PR #105, opened 2016-03-27
Loot

madmouser1:
Initial Loot.co.za importer

### DIFF
--- a/loot_importer.user.js
+++ b/loot_importer.user.js
@@ -33,6 +33,7 @@ if (DEBUG) {
  * - http://www.loot.co.za/product/various-artists-30-30-goue-sokkie-treffers-volume-17/hbhf-3088-g440
  * - http://www.loot.co.za/product/jacques-de-coning-veels-geluk/lzmg-572-g610
  * - http://www.loot.co.za/product/various-artists-vat-5-volume-5/dfnc-3405-g1a0
+ * - http://www.loot.co.za/product/bette-midler-a-gift-of-love/mhgm-3483-g060  *** NOT WORKING *** extra tab
  */
 
 
@@ -59,6 +60,13 @@ function insertMbUI(mbUI) {
   }
 }
 
+// Insert links to high res image in Loot page
+function insertIMGlinks() {
+  var imghref = $('#imagePreview0 a.fancybox').attr('href');
+  imghref = 'http://static.loot.co.za/' + imghref;
+  LOGGER.debug("insertIMGlink Firing", imghref);
+  $('#imagePreview0').append('<p><a href="' + imghref + '">MB High Res Image</a></p>');
+}
 
 // Insert links in Loot page
 function insertMBSection(release) {
@@ -73,7 +81,7 @@ function insertMBSection(release) {
   mbUI.append(mbContentBlock);
 
   if (release.maybe_buggy) {
-    var warning_buggy = $('<p><small><b>Warning</b>: this release has perhaps a buggy tracklist, please check twice the data you import.</small><p').css({
+    var warning_buggy = $('<p><small><b>Warning</b>: this release has perhaps a buggy title, please check twice the data you import.</small><p').css({
       'color': 'red',
       'margin-top': '4px',
       'margin-bottom': '4px'
@@ -91,6 +99,7 @@ function insertMBSection(release) {
   mbContentBlock.append(innerHTML);
 
   insertMbUI(mbUI);
+  insertIMGlinks();
 
   $('#mb_buttons').css({
     display: 'inline-block',
@@ -110,7 +119,6 @@ function insertMBSection(release) {
 
   mbUI.slideDown();
 }
-
 
 function parseReleaseDate(rdate) {
   var months = {
@@ -137,9 +145,6 @@ function parseReleaseDate(rdate) {
   }
   return false;
 }
-
-
-
 // Analyze Loot data and return a release object
 function ParseLootPage() {
   LOGGER.debug("ParseLootPage function firing");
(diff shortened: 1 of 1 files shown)


loot_importer.user.js
```diff
@@ -0,0 +1,574 @@
...
+      LOGGER.debug("<TD>", prodinfolabellowcase);
+
+      switch (prodinfolabellowcase) {
+        case "label:": // use these cases to select the spesific text values
+          LOGGER.debug("* label: selected *");
+          var release_label = new Object();
+          release_label.name = prodinfolabel.nextElementSibling.innerText.trim();
+          prodlabels.push(release_label);
+          break;
+        case "release date:": // "release date:" TODO: NOT WORKING
```
madmouser1: Please assist as case "release date:" is never matched and I can't see the fault

zas: Since we include jquery you should use it instead, something like:

```javascript
// extract all tr from table with class productDetails
$("table.productDetails tr").each(function() {
 // get text from first td, trim and convert it to lowercase
  var prodinfolabellowcase = $(this).children('td').eq(0).text().trim().toLowerCase();
  // get test from second td, which is the corresponding value
  var value = $(this).children('td').eq(1).text().trim();
  // now compare and process
  switch (prodinfolabellowcase) {
          case "label:": // use these cases to select the spesific text values
          prodlabels.push({name: value});
          break;
  /* etc... */
  }
 
});
```

To parse month and year you can use something like https://github.com/murdos/musicbrainz-userscripts/blob/6db55d24942d69580794899b5b2c196fae229a63/juno_download_importer.user.js#L29

loot_importer.user.js
```diff
@@ -126,50 +156,70 @@ function ParseLootPage() {
...
+      case "performers:":
+        LOGGER.debug("** performers: **", value);
+        release_artist_array.push({
+          name: value
+        });
+        break;
+      case "format:":
+        LOGGER.debug("* format: **");
+        break;
+        /* etc... */
```
zas: No point in this comment that was in my example ;)

loot_importer.user.js
```diff
@@ -126,50 +156,70 @@ function ParseLootPage() {
...
+
+
+  LOGGER.debug("Release Title:", releasetitle, "  Release Artist:", releaseartist, "  Release Format:", release_format);
+
+
+  // extract all tr from table with class productDetails
+  $("table.productDetails tr").each(function() {
+    // get text from first td, trim and convert it to lowercase
+    var prodinfolabellowcase = $(this).children('td').eq(0).text().trim().toLowerCase();
+    prodinfolabellowcase = prodinfolabellowcase.replace(/\s+/g, ''); //removing white spaces as switch isnt matching spaces for some reason
```
zas: Huh? `switch()` should have no issue matching strings with spaces...
However, i think you can simply use `if` / `else if` here, it will not make much difference.

loot_importer.user.js
```diff
@@ -126,50 +156,70 @@ function ParseLootPage() {
...
   // div#productContent table tbody tr.productOverview td.productInfo h1
   var AlbumName = document.querySelectorAll("div#productContent > table > tbody > tr.productOverview > td.productInfo > *");
-  LOGGER.debug("Album Name: ", AlbumName[0].innerText); // textContent give the space formatting and we just need the text so we use innerText
-  releasetitle = AlbumName[0].innerText;
-  LOGGER.debug("Artist Name: ", AlbumName[1].innerText);
+
   releaseartist = AlbumName[1].innerText;
+  if (releaseartist == "Various Artists") {
+    // Everything is: title(format)
+    releaseartisttitle_regex = /(.*)\((.*)\)/;
```
zas: I would use to ensure we match external parenthesis:
```
releaseartisttitle_regex = /(.*?)\((.*)\)/;
```

Compare:
```javascript
var s = 'A(B(C))';
var r = /(.*)\((.*)\)/;
var m = str.match(r);
console.log(m);
```
`["A(B(C))", "A(B", "C)"]`

To:

```javascript
var s = 'A(B(C))';
var r = /(.*?)\((.*)\)/;
var m = str.match(r);
console.log(m);
```

`["A(B(C))", "A", "B(C)"]`

loot_importer.user.js
```diff
@@ -126,50 +156,70 @@ function ParseLootPage() {
...
+    releaseartisttitle = AlbumName[0].innerText.match(releaseartisttitle_regex);
```
zas: What if match fails ? `releaseartisttitle` can be `null` (for example, if matched text is `A(B`)

loot_importer.user.js
```diff
@@ -126,50 +156,70 @@ function ParseLootPage() {
...
+
+
+    releasetitle = releaseartisttitle[1];
+    release_format = releaseartisttitle[2];
+
+  } else {
+    // 
+    releaseartisttitle_regex = /(.*) - (.*)\((.*)\)/;
```
zas: See previous comments

loot_importer.user.js
```diff
@@ -190,13 +240,14 @@ function ParseLootPage() {
       descriptiontrack = new Object();
 
       var currenttrack = tracklisting[trackiterate].querySelectorAll("td");
-      var artisttitle_regex = /(.*) - (.*)/; // regex: artist - title 
+      //   var artisttitle_regex = /(.*) - (.*)/; // regex: artist - title 
+      var artisttitle_regex = /(.*) (-|–) (.*)/; // regex: artist - title char 45 or 8211
```
zas: Perhaps you should do the same in previous regexs ?

loot_importer.user.js
madmouser1: I have no idea why the switch() don't match the string with spaces in this instance, I can only think that it is not a true space but some other HTML space character

PR #109, opened 2016-04-08
TakeAlot importer: added checks and format parser

madmouser1:
added release format & label check from product info tab
created div format parser for description tab
assume missing artist if not present and mark release object as buggy
insert high res image link on page


zas:
That's a lot of changes in one commit, not easy to review, can you **split it in smaller commits** please?

Also think about prefixing the PR title with the importer, it helps to know on what it applies, as we have many importers in the same project.

Can you explain the need of `removeDiacritics()` ?

madmouser1:
removeDiacritics() : was added to get around a non-standard naming convention they use.
I compare the artist name in the heading with the artist name in the product details, if the same all is ok else I need to swap artist and title in the heading.

But on samples I used Takealot had diacritics in the Artist name in the heading but not in the product details, hence me removing it in the heading to compare like for like.

PR #111, opened 2016-05-07
FMA: initial import seeder

madmouser1:
This is the initial script for testing and comments please;
It uses the FMA API (http://freemusicarchive.org/api/docs/) rather than page information to build the release object.

Script work as follows:

- Check if API key is set for registered user
- The first time you enter your API KEY it will store it with localStorage.
- Parse data that is only obtainable from the website (including the album id needed for API & release label)
- Do all the AJAX calls to retrieve the data from API (The FMA API limit results to 20 per call)
- Once all the data retrieved a release object is build and presented.

- UI: added some on-page indicators to track API call status during debug.

Debug set to true during testing phase

### DIFF
--- a/fma_importer.user.js
+++ b/fma_importer.user.js
@@ -34,7 +34,7 @@ var FMA_API = "FMA API KEY Missing";
 
 if (!unsafeWindow) unsafeWindow = window;
 
-var DEBUG = true; // true | false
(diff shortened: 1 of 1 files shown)


madmouser1:
@zas you may merge

PR #126, opened 2016-10-28
dgdecca_importer

loujine:
New importer from Deutsche Grammophon/Decca website

Adapted from https://greasyfork.org/en/scripts/24375-import-dg-decca-releases-to-musicbrainz in order to use the same downloadURL/icon scheme as others in murdos/musicbrainz-userscripts

The userscript probably needs more comments, but at least it has been working well for me


zas:
Thank you :+1:

PR #131, opened 2016-12-04
Merge facebook-importer userscript

webhaikal:
GCI Task


zas:
Please rebase this PR 14 commits without message + 1 merge commit .... Do a proper rebase to clean up this mess.

zas:
Since the script imports facebook events and doesn't use any of the code of other importers, it would be better to have it named `facebook_events_importer.js`

facebook_importer.user.js
```diff
@@ -0,0 +1,74 @@
...
+// @name        MusicBrainz: Import events from Facebook
+// @version     2016-01-17
+// @author      Ohm Patel
+// @require     https://code.jquery.com/jquery-2.2.0.min.js
+// @grant       none
+// @include     *.facebook.*/events/*
+// ==/UserScript==
+
+var API_URL = "//graph.facebook.com/v2.5/";
+var API_KEY = "CAAO2IKDySlMBAOc6UU5jRHqiEFrENsZC09f4KJsMpDPOTV9gAjZC5tIdVqBeibxfk0i2XwuHL61ywnQdnI4KdCgwCfZCXwc0ZCryj48a1PRjVMuCbpoScvzeYlrTujLynPPKdgS66eLAvqGZCzl3Kj7ahRj97tYeR4VIGGVCQ9pRWDxZC9rPzIBEFTjvjBZCacZD";
```
Freso: I think we had some discussion about this when @opatel99 wrote the script in January. @opatel99, can you remember what's going on here?

Freso:
The point of the exercise was to import the userscript with the commit history from its old location, https://gist.github.com/opatel99/3537237628abce4da2a3 - I would hate to lose the original change history of the script, but I would also want to get it put into a "proper" repository. :/

I already made a comment on the GCI task to rename the file, but your name suggestion is better. :) (I just asked to rename from using "-" to "_".)

zas:
> The point of the exercise was to import the userscript with the commit history from its old location, https://gist.github.com/opatel99/3537237628abce4da2a3 - I would hate to lose the original change history of the script, but I would also want to get it put into a "proper" repository. :/

Then it will be better to rebase AND add messages to commits, it would be an even better exercise ;)

I didn't see it was coming from a gist.

Freso:
I don't think it's fair for @webhaikal to understand @opatel99's commits. :) But once @webhaikal has renamed the script again, I'll approve the GCI task and we'll see if maybe @opatel99 can take over the PR and add messages to his commits, if he can remember what/how/why he changed.

webhaikal:
@Freso, I have renamed the file.

README.md
```diff
@@ -139,6 +139,13 @@ Add a button to import Takealot releases to MusicBrainz
 [![Source](https://github.com/jerone/UserScripts/blob/master/_resources/Source-button.png)](https://github.com/murdos/musicbrainz-userscripts/blob/master/takealot_importer.user.js)
 [![Install](https://raw.github.com/jerone/UserScripts/master/_resources/Install-button.png)](https://raw.github.com/murdos/musicbrainz-userscripts/master/takealot_importer.user.js)
 
+## <a name="facebook_importer"></a> Import Facebook Events to MusicBrainz
```
Freso: The `name` should be `facebook_events_importer` (to be consistent with the rest).

webhaikal:
@Freso it should be fixed now :smile:

README.md
```diff
@@ -1,5 +1,6 @@
 # MusicBrainz UserScripts
 
+* [Import Facebook events to MusicBrainz](#facebook_events_importer)
```
Freso: It looks like the list is alphabetically sorted, so this entry should be at the top of list. ;)

README.md
```diff
@@ -20,6 +21,13 @@
 * [Musicbrainz DiscIds Detector](#mb_discids_detector)
 * [Musicbrainz UI enhancements](#mb_ui_enhancements)
 
+## <a name="facebook_events_importer"></a> Import Facebook Events to MusicBrainz
+
+Add a button to import events from Facebook to MusicBrainz
+
+[![Source](https://github.com/jerone/UserScripts/blob/master/_resources/Source-button.png)](https://github.com/murdos/musicbrainz-userscripts/blob/master/facebook_events_importer.js)
+[![Install](https://raw.github.com/jerone/UserScripts/master/_resources/Install-button.png)](https://raw.github.com/murdos/musicbrainz-userscripts/master/facebook_events_importer.js)
```
Freso: As the other comment: the list seems to be sorted alphabetically, so this entry shouldn't be at the top. :)

webhaikal:
@Freso is this right?

README.md
zas: Please do not change README.md in PR, it will be (re-)generated after the merge.

facebook_events_importer.js
```diff
@@ -0,0 +1,74 @@
+// ==UserScript==
+// @name        MusicBrainz: Import events from Facebook
+// @version     2016-01-17
+// @author      Ohm Patel
+// @require     https://code.jquery.com/jquery-2.2.0.min.js
```
zas: Please add headers as in https://github.com/murdos/musicbrainz-userscripts/blob/master/bandcamp_importer.user.js#L1-L15

Check version format: YYYY.MM.DD.N
Add `@downloadURL`, `@updateURL` and `@icon` headers
Those URLs and correct version are required to allow automatic upgrades.

Adapt `@author`.

Add `// @namespace      https://github.com/murdos/musicbrainz-userscripts`

And a proper description (`@description`)

facebook_events_importer.js
```diff
@@ -0,0 +1,74 @@
...
+// @grant       none
+// @include     *.facebook.*/events/*
```
zas: I would match protocol too: `http*://*.facebook.*/events/*`
Also this expression may match `https://www.facebook.com/anything/something/anotherthing/events/things/stuff`, you may want to use stricter regex as shown at https://github.com/murdos/musicbrainz-userscripts/blob/master/bandcamp_importer.user.js#L8 for example.

README.md
Freso: Oh, okay. @webhaikal – see what @zas said and forget everything I've said about the README. :)

facebook_events_importer.js
```diff
@@ -0,0 +1,74 @@
...
+    add_field("edit-event.name", eventData.name);
+    add_field("edit-event.edit_note", "Imported from " + window.location.href + " using MusicBrainz - Facebook Event Importer");
+    
+    add_field("edit-event.url.0.text", window.location.href);
+    add_field("edit-event.url.0.link_type_id", "783");
+    
+    add_field("edit-event.period.begin_date.year", eventData.start_time.substring(0,4));
+    add_field("edit-event.period.begin_date.month", eventData.start_time.substring(5,7));
+    add_field("edit-event.period.begin_date.day", eventData.start_time.substring(8,10));
+    add_field("edit-event.time", eventData.start_time.substring(11,16));
```
zas: Imho, stricter parsing is needed.
Also timezone is ignored, FB API is returning events datetime in following format, with timezone set to the one of the user (token):
` ""end_time": "2017-01-14T23:30:00+0100",`

```javacript
var d = new Date(Date.parse('2017-01-14T23:30:03+0500'));
function convertDateToUTC(date) {
  return [date.getUTCFullYear(), date.getUTCMonth()+1, date.getUTCDate(), date.getUTCHours(), date.getUTCMinutes(), date.getUTCSeconds()];
}

convertDateToUTC(d);
```

will output an array representing the UTC datetime:
```[2017, 1, 14, 18, 30, 3]```

zas:
> Okay, so apparently the README should not be altered… sorry about that!

Well, it isn't obvious it is autogenerated, may be i need to add a line in it about that.

facebook_events_importer.js
webhaikal: I don't understand this part @zas 

var d = new Date(Date.parse('2017-01-14T23:30:03+0500'));
function convertDateToUTC(date) {
  return [date.getUTCFullYear(), date.getUTCMonth()+1, date.getUTCDate(), date.getUTCHours(), date.getUTCMinutes(), date.getUTCSeconds()];
}

convertDateToUTC(d);

zas:
> I don't understand this part @zas

Well, apparently FB API is returning date+time with a timezone information, so you need to convert the date and time to UTC timezone so it doesn't depend on user timezone. Perhaps there's a mean to get time & data in Unix timestamp format (see https://chris.banes.me/2011/06/24/correctly-parsing-graph-api-event-times/)

So, to be short, be sure the date & time you import to MB is in the correct format.

Freso:
@webhaikal Note that @zas' code cleanup comments are not necessary for completing the (current) GCI task.

There will be another GCI task for cleaning up the script afterwards, which I'd be happy to have you claim. :)

@zas We want the time in the event's local time in MB though, not in UTC (and not in the user's timezone—unless the user is in the same timezone as the event). I don't know if this is doable somehow. :)

zas:
> We want the time in the event's local time in MB though

Huh, it looks a bad idea to store time & date in anything but UTC, but i didn't find actual docs about importing events in MB.
It doesn't change it shouldn't depend on FB user timezone.

webhaikal:
@Freso sorry about that. Renamed the commit for something better :wink:

PR #144, opened 2017-04-16
Fix #136: mb_relationship_shortcuts.user.js not handling GET parameterrs properly

### LINKED ISSUE #136: Display shortcut for relationships on MusicBrainz: does not work for multipage lists
When browsing entities that have more than 100 entries, those are split up.

On the first view, relationships are displayed correctly:
https://musicbrainz.org/label/672a919c-9050-479a-827c-5570deeaa0c3

If going to page 2, 3, or back to page 1, this script does not show relationships anymore:
https://musicbrainz.org/label/672a919c-9050-479a-827c-5570deeaa0c3?page=1

The reason is probably the appended "?page= ".

LeoVerto:
If a release group or label URL included additional GET parameters, namely ?page=X when switching to another page, the script regex would not pick it up anymore.

### DIFF
--- a/mb_relationship_shortcuts.user.js
+++ b/mb_relationship_shortcuts.user.js
@@ -55,7 +55,7 @@ $(document).ready(function(){
         parent.type = 'artist';
         parent.mbid = m[1];
         child.type = 'release-group';
-    } else if (m = window.location.href.match("\/(release-group|label)\/(.{36})$")) {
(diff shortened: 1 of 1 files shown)


LeoVerto:
I haven't bumped the userscript version yet, should that be done as part of this PR?

zas:
> I haven't bumped the userscript version yet, should that be done as part of this PR?

Better not, since it may conflict with other pending PRs (not in this case).
I usually update versions using tools/update_version.sh script after the merge.

PR #155, opened 2017-09-09
Beatport empty Playable bugfix

Ruj89:


beatport_importer.user.js
```diff
@@ -62,6 +62,10 @@ function retrieveReleaseInfo(release_url) {
       catno: ProductDetail.catalog
     }
   );
+  
+  // Reload Playables if empty
+  if(!unsafeWindow.Playables.hasOwnProperty("tracks"))
+    eval($("#data-objects").text());
```
roman412: Just evaluating this script again is not enough, because we copied `window` into `unsafeWindow` and evaluating updates `window` object but not `unsafeWindow`.

```js
if(!unsafeWindow.Playables.hasOwnProperty("tracks")) {
    eval($("#data-objects").text());
    unsafeWindow.Playables = window.Playables;
}
```
This can solve the issue.

Ruj89:
OK, i fixed the branch. Really thanks for the help!

roman412:
Forgive me being picky, but maybe you can squash your two commits into one?
These two commits are the same piece of code that was finalized, I think that there's no need to be split into two commits. Also, my messages about changes are superfluos as the changes are applied.

Thank you for fixing this.

Ruj89:
Squashed

roman412:
@zas could you review, please?

roman412:
> zas closed this 2 days ago

Why not?

zas:
This PR was stale since more than one year, I closed it by error.
Since the source repo was removed since, I can't just re-open it, so I re-created it as https://github.com/murdos/musicbrainz-userscripts/pull/244

PR #161, opened 2017-11-11
Display more relationships + minor changes

SothoTalKer:
- updated the script so that a few "other databases" relationships are shown aswell; might be kludgy but better than not showing them at all
- changed most of the favicons to use the copies stored on the MusicBrainz server
- remove unused logger, updated to latest jQuery without any noticeable breakage
- fixed some minor code warnings

### DIFF
--- a/mb_relationship_shortcuts.user.js
+++ b/mb_relationship_shortcuts.user.js
@@ -1,7 +1,7 @@
 // ==UserScript==
 // @name           Display shortcut for relationships on MusicBrainz
 // @description    Display icon shortcut for relationships of release-group, release, recording and work: e.g. Amazon, Discogs, Wikipedia, ... links. This allows to access some relationships without opening the entity page.
-// @version        2017.11.11.0
+// @version        2017.11.12.0
 // @author         Aurelien Mino <aurelien.mino@gmail.com>
 // @licence        GPL (http://www.gnu.org/copyleft/gpl.html)
 // @downloadURL    https://raw.github.com/murdos/musicbrainz-userscripts/master/mb_relationship_shortcuts.user.js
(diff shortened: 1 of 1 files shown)


mb_relationship_shortcuts.user.js
```diff
@@ -117,11 +116,24 @@ $(document).ready(function(){
             $(this).find("relation-list[target-type='url'] relation").each(function() {
                 var reltype = $(this).attr("type");
                 var target = $(this).children("target").text();
-                if (relationsIconsURLs['url'].hasOwnProperty(reltype)) {
+                if (relationsIconsURLs.url.hasOwnProperty(reltype)) {
+                    $("#" + mbid + " td.relationships").append(
+                        "<a href='" + target.replace(/'/g,"&apos;") + "'>" + "<img style='max-height: 16px;' src='" + relationsIconsURLs.url[reltype] + "' />&nbsp;" + "</a>"
+                    );
+                }
+                if (target.indexOf("d-nb.info") != -1 ) {
```
murdos: Please declare a global dictionary in the start of the script, like relationsIconsURLs, with all additional sites and their icons.
It's better to separate data and logic processing. This will also avoid duplicating the same code block for each additional site.

murdos:
Thanks for cleaning and updating this old code!
Once you have take into account my main concern (dictionary of other databases relationships), I'll accept this merge ;-)

mb_relationship_shortcuts.user.js
SothoTalKer: I will have a look if my skills allow for this. I can't promise anything, though.

SothoTalKer:
I think this should be fine. Data and logic are separated now, and no more code duplication. :)

murdos:
Thanks!

PR #165, opened 2017-12-30
Check target type in shorcut for relationships

yvanzo:
Issue: `mb_relationship_shortcuts.user.js` triggers the following exception under certain circumstances:

````
TypeError
relationsIconsURLs[targettype] is undefined
````

This patch tests if the `targettype` is appropriate for relationships shortcut display.

### CODE REVIEW

**Comment by zas** [comment]
@yvanzo : update the version so the userscript will be auto upgraded (for those having this option)

```bash
./tools/update_version.py  mb_relationship_shortcuts.user.js
```

yvanzo:
@zas : script tested, version bumped

murdos:
Merged. Thanks yvanzo!

PR #169, opened 2018-02-07
URGENT: Recording comment removal edits are queued!

jesus2099:
Script does no longer fetch recording comments for edit.
It’s ending up in submitting recording comment removal edits!
Because of `_.pluck is not a function` issue.

---

I have also cleaned up <del>`@include`</del>, `@exclude` and `@match`.

### DIFF
--- a/set-recording-comments.user.js
+++ b/set-recording-comments.user.js
@@ -1,7 +1,7 @@
 // ==UserScript==
 // @name           MusicBrainz: Set recording comments for a release
 // @description    Batch set recording comments from a Release page.
-// @version        2016.5.30
+// @version        2018.2.17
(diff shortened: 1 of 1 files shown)


jesus2099:
Thanks very much @zas.
I hope it will always work because I don’t know anything about underscore.js.
It’s a quick (and dirty?) fix to avoid unintended removal edits like https://musicbrainz.org/edit/50779852 .

jesus2099:
I’m currently looking for buggy edits and found https://musicbrainz.org/edit/50747466 but it’s difficult to search.
I think I should add something to prevent edits if there is any error.
And maybe a script edit note signature to allow retrieving bug edits more easily.

PR #171, opened 2018-02-17
Fix update version to make sub-version numbers without leading 0

yvanzo:
As reported by @jesus2099 in https://github.com/murdos/musicbrainz-userscripts/commit/b365af8cde4117cd65bb2f8274b5c0bb7e83d9a8#commitcomment-27623531 and https://github.com/murdos/musicbrainz-userscripts/commit/b365af8cde4117cd65bb2f8274b5c0bb7e83d9a8#commitcomment-27605542, version part numbers starting with `0` are not supported by Chrome at least.

This patch fixes the tool to update version so as to stop padding month and day with leading `0` and to start the fourth number/part of version from `1` instead of `0`.

As ane example, it also bumps version number for recently updated `mb_relationship_shortcuts.user.js` which caused this bug to be reported.

If this patch does not generate other any auto update bug for other platform/browser, it should be fine to bump the version number of every other script with such unsupported version number.

### DIFF
--- a/mb_relationship_shortcuts.user.js
+++ b/mb_relationship_shortcuts.user.js
@@ -1,7 +1,7 @@
 // ==UserScript==
 // @name           Display shortcut for relationships on MusicBrainz
 // @description    Display icon shortcut for relationships of release-group, release, recording and work: e.g. Amazon, Discogs, Wikipedia, ... links. This allows to access some relationships without opening the entity page.
-// @version        2018.02.11.0
+// @version        2018.2.17.1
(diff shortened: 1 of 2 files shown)


zas:
Ok, it is a workaround for an unfixed chrome issue. But how such version change will impact other userscript engines ? I guess it will force an upgrade of all scripts (for people having autoupgrade feature enabled), and therefore isn't much an issue. Is it right?

yvanzo:
Yes. `2` will probably be considered to be the same or higher than `02`, not lesser.

I tested the script that updates number only, which considers `02` and `2` to be the same.

PR #177, opened 2018-02-28
Discog importer: Turn all acute accents used as apostrophe into the proper character

jesus2099:
Acute accents are wrong and they also make the MB track parser add spaces around it.
Example https://www.discogs.com/release/7809058 → https://musicbrainz.org/edit/51145485
This is only small change proposal on track titles but if wanted,
it should be done for track artists and release title and artists as well.
Maybe as a new cleanup function.
Tell me.


jesus2099:
I don’t think there should be many such cases of discogs with ACUTE instead of apostrophe so I don’t think the script should even bother, well, really no opinion…

PR #184, opened 2018-06-01
Bandcamp Importer: Use only https URLs

zas:
- Bandcamp now redirects all http URLs to https
- They didn't change URLs in DOM data yet, most are still http
- MB changed Bandcamp URLs normalization to use https
- MB didn't change old Bandcamp http URLs in the database yet

Basically this breaks some features of the importer, like detection of existing artists, labels, and releases (small icons)
Since the importer has no way to know if the url is stored as http or https, the result is quite random.

One solution would be to query for both, doubling the number of requests to the web service.
Another, much better, is to convert all old Bandcamp http URLs in MB database to https (all new URLs are normalized to https).

This patch only fix http URLs still appearing in pages, and look up for https URLS on the web service.
So, at the moment, the fix is incomplete.

### DIFF
--- a/bandcamp_importer.user.js
+++ b/bandcamp_importer.user.js
@@ -1,7 +1,7 @@
 // ==UserScript==
 // @name           Import Bandcamp releases to MusicBrainz
 // @description    Add a button on Bandcamp's album pages to open MusicBrainz release editor with pre-filled data for the selected release
-// @version        2018.3.7.1
+// @version        2018.6.1.1
 // @namespace      http://userscripts.org/users/22504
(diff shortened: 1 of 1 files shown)


murdos:
+1

yvanzo:
> to convert all old Bandcamp http URLs in MB database

It is already in progress, using the conversion script in https://github.com/metabrainz/musicbrainz-server/pull/677. So far, 12K URLs have been converted, there still are 132K URLs to be converted. It will take a few days to complete.

zas:
The whole conversion should be finished today.

PR #194, opened 2018-11-12
Init package.json + setup prettier and eslint

murdos:
Setup prettier to unify formatting of source code, and eslint to enforce rules (and automatically fix some)

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,14 @@
+# Editor configuration, see http://editorconfig.org
+root = true
+
+[*]
+charset = utf-8
+indent_style = space
+indent_size = 4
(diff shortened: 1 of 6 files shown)


murdos:
Please let me know if you agree on the proposed configuration, before applying it on existing code.

murdos:
>
> Please be sure to update versions of scripts after prettyfication, i had a quick look over changes and it seems ok, but there are a lot of changes.

I was initially not sure if a publication of new versions was necessary, but I think you're right.
I however plan to fix manually a few eslint rules that I temporarily set to 'warn', after the wave of automatic changes. I'll update versions of scripts subsequently.

PR #195, opened 2018-11-15
Qobuz fixes

Jormangeud:


murdos:
@zas since you're the author of this script, could you review this PR before we apply all format changes induced by #194 ? Thanks ;-)

qobuz_importer.user.js
```diff
@@ -42,10 +42,10 @@ function parseRelease(data) {
 
   // Release information global to all Beatport releases
   release.packaging = 'None';
-  release.country = "";
+  release.barcode = data.upc;
+  release.country = "XW";
   if (i18n_global && i18n_global.zone) {
-    if (i18n_global.zone == 'GB') release.country = 'UK';
```
zas: I think this line should stay, it converts 'GB' used by Qobuz to 'UK' used by MB

qobuz_importer.user.js
```diff
@@ -133,30 +135,27 @@ function insertLink(release) {
   mbUI.slideDown();
 }
 
+// Hook all XMLHttpRequest to use the data fetched by the official web-player.
+(function() {
+    const send = XMLHttpRequest.prototype.send
+    XMLHttpRequest.prototype.send = function() {
+        this.addEventListener('load', function() {
+            var wsUrl = 'https://www.qobuz.com/api.json/0.2/album/get?album_id=';
```
zas: Why is `app_id` removed ?

Jormangeud: We do not need it, as the new code piggybacks on web player. There were two exactly same requests for the album data with different app_ids. I thought that hooking to the first request saves one of those, and wont need a new app_id.

I think the current app_id has been deactivated, as the request was denied (that was reported in the issue).

qobuz_importer.user.js
Jormangeud: That is to say, I use the site in en-gb and GB is valid for seeding, but UK isn't.

qobuz_importer.user.js
```diff
@@ -119,8 +129,9 @@ function insertLink(release) {
   var parameters = MBImport.buildFormParameters(release, edit_note);
 
   var mbUI = $('<p class="musicbrainz-import">' + MBImport.buildFormHTML(parameters) + MBImport.buildSearchButton(release) + '</p>').hide();
+  var isrcslist = $('<textarea>' + release.isrcs.join("\n") + '</textarea>');
```
zas: That's valuable information but not anyone may be interested in ISRCs, i'd rather provide a way to show/hide this textarea.
Something like http://jsfiddle.net/Ab9vX/3/ could do.
And this option should only be possible if the textarea isn't empty (i don't think all albums on Qobuz have ISRCs).

PR #197, opened 2018-11-20
Code cleanup

murdos:


murdos:
Yeah, the current changes are good, but my intent was to continue with more cleanup.
But since I'm busy with other things, we should merge this.

PR #198, opened 2018-11-21
qobuz_importer.user.js: Support for classical releases.

Jormangeud:
This will use composer as artist on various artists/composers releases.
Refactored code to functions as using composers as well as artists.
Added function for guessing various artists, as the hardcoded are
likely label based and are hard to keep updated.

Also, fixing a bug in single artist use case on a track.


qobuz_importer.user.js
```diff
@@ -26,19 +26,50 @@ if (DEBUG) {
...
 
     release.script = 'Latn';
     release.url = `https://www.qobuz.com${data.relative_url}`; // no lang
 
     release.title = data.title;
-    release.artist_credit = MBImport.makeArtistCredits([data.artist.name]); // FIXME: various artists
-    if ($.inArray(data.artist.id, various_artists_ids) != -1) {
-        release.artist_credit = [MBImport.specialArtist('various_artists')];
-    }
+    if (isVariousArtists(data.artist)) {
```
zas: So it only works for Various Artists albums having "Classique" as genre ?
What about non-french Qobuz websites ? And non-various artists classical albums ?

https://www.qobuz.com/de-de/album/brahms-piano-quintet-piano-quartets-nos-1-3-quatuor-hermes-and-geoffroy-couteau/rwp9d27ev3ytb

I wonder if having a checkbox allowing the user to decide wouldn't be better (and the checkbox can be checked by auto detectiion).

Jormangeud: Yes, it's using composer for artist just in classical various releases now. It would probably be better to detect the classical before various, and use composer when artist is missing. As with the link you gave, there is no artist, only performer and composer, so its' artist is the ensemble.

Classique seems to be constant in the genres_list from the API, so it is working at least in gb-en, fr-fr and de-de. There is store_related object with translations. And trying some of those store locales, some work, some require to be registered to work, some give 404 (seem that there are some georestrictions both to releases and stores). 

The checkbox with auto detection idea is great. At the moment the parsing to the mb release is done on load, and to be able to parse it again, the data object needs to be stored.

I think I'll implement both of those.

qobuz_importer.user.js
```diff
@@ -132,13 +185,28 @@ function insertLink(release) {
...
+        .hide()
+        .append($('<button id="isrcs" type="submit" title="Show list of ISRCs">Show ISRCs</button>'));
+    if (classical) {
+        let title = 'Release data was detected as classical. Click to parse as artist.';
+        if (!detect) title = 'Click to reparse as classical.';
+        mbUI.append($(`<button id="reparse" type="submit" title="${title}">Reparse</button>`));
+        mbUI.append(
+            $(
+                `<span><input id="force_aa" type="checkbox" title="Force album artist"${force_aa &&
+                    ' checked'}><label for="force_aa">Force album artist</label></span>`
```
zas: It'd be better if the checkbox and the label were on the same line (Firefox here).
image

Also tooltips aren't very clear about what it changes imho.
If it is possible we should just propose this choice if it is actually changing something.

Do we really need the "Reparse" option ? I mean what about doing parsing once, and keep both cases, if they differ.

Looks promising :)

qobuz_importer.user.js
```diff
@@ -197,3 +265,17 @@ $(document).on('click', '#isrcs', function() {
     } else $(this).text('Show ISRCs');
     return false;
 });
+
+$(document).on('click', '#reparse', function() {
+    detect = !detect;
```
zas: Imho, `classical_detected` would be better than just `detect`

qobuz_importer.user.js
```diff
@@ -197,3 +265,17 @@ $(document).on('click', '#isrcs', function() {
...
+    let release = parseRelease(raw_release_data);
+    insertLink(release);
+    return false;
+});
+
+$(document).on('click', '#force_aa', function() {
+    force_aa = !force_aa;
```
zas: `force_aa` -> `force_album_artist` ?

PR #205, opened 2019-01-05
Remove unsafeWindow usage in logger and in scripts that don't need it

murdos:
and grant unsafeWindow for scripts that actually use it.


murdos:
Indeed. 
I'll fix the conflict and merge this PR.

PR #213, opened 2019-03-18
Friendlier make edit note

brianfreud:
Moved 'home' to arguments as an optional parameter with a default value, to allow it to be changed for scripts that don't live at murdos' github.  

Also fixed a few 'var's that should have been 'let's - 'i' wasn't scoped, so the 2nd, 3rd, and 4th 'var i' statements were attempting to redeclare an already existing 'i'.  Now they're all scoped to just the loop.

### DIFF
--- a/lib/mbimport.js
+++ b/lib/mbimport.js
@@ -157,7 +157,7 @@ var MBImport = (function() {
         buildArtistCreditsFormParameters(parameters, '', release.artist_credit);
 
         if (release['secondary_types']) {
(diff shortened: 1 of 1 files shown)


brianfreud:
Sounds good.  Ty - and quite a nice import processor.  Saved me having to go back and overhaul my old one. :)

I didn't dig too much to figure out quite what the format argument is used for, but it looks like the code might also be simplified a little by moving the ternary into a default '' value for that argument.
ie this bit --> ${format ? ` (${format})` : ''}

Issue #220, opened 2019-04-11
Discogs Memory Stick should be imported as USB Flash Drive


jesus2099:
[This release](https://musicbrainz.org/edit/60615348) where a [Memory Stick](https://www.discogs.com/release/12679820) was imported as **Other** instead of as [USB Flash Drive](https://musicbrainz.org/edit/60619278).

liftarn:
Fixed it in https://github.com/murdos/musicbrainz-userscripts/pull/221

PR #221, opened 2019-04-11
Discogs Memory Stick should be imported as USB Flash Drive #220

liftarn:


liftarn:
@murdos Merge this please.

PR #231, opened 2019-06-24
New userscript: MusicBrainz: 1200px CAA

Freso:
This script makes the release/*/cover-art pages use the 1200px “thumbnail” images for the pop‐up/preview instead of the original image, which can speed up loading times significantly for large scans, but should still be high enough definition for most text to be legible.

Additionally it adds a “1200px” link to the “All sizes” lists. This will likely be unneeded as soon as [CAA-88](https://tickets.metabrainz.org/browse/CAA-88) is resolved, but until then it is really nifty to have handy.

Note: A few (older) images may not be available in 1200px.

---

Adding as a PR in case anyone objects to it going into this repository. :) (I’m using it myself right now, and it’s working wonderfully!)

### DIFF
--- a/mb_1200px_caa.user.js
+++ b/mb_1200px_caa.user.js
@@ -1,0 +1,42 @@
+// ==UserScript==
+// @name           MusicBrainz: 1200px CAA
+// @name:da        MusicBrainz: 1200px CAA
+// @namespace      https://github.com/murdos/musicbrainz-userscripts/
+// @version        2019.6.24.1
+// @author         Frederik “Freso” S. Olesen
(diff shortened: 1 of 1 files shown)


murdos:
No objection to add this userscript 😊

murdos:
However I would not have merged it with the errors reported by the linter...

zas:
> However I would not have merged it with the errors reported by the linter...

I merged it, because i wanted to update README.md too (and the script used to generate it).
I fixed "errors" (mostly stylistic) in https://github.com/murdos/musicbrainz-userscripts/pull/232

PR #232, opened 2019-06-25
Make MusicBrainz: 1200px CAA prettier

zas:


mb_1200px_caa.user.js
```diff
@@ -21,22 +21,22 @@ var ca_page = document.querySelector('div#content');
 var ca_items = ca_page.querySelectorAll('div.artwork-cont');
 
 ca_items.forEach(function(ca_item) {
-  /* Use 1200px “thumbnails” for the pop‐ups/previews */
-  var popup_link = ca_item.querySelector('a.artwork-image');
-  popup_link.href = popup_link.href.replace(/\.[a-z]+$/, '-1200.jpg');
+    /* Use 1200px “thumbnails” for the pop‐ups/previews */
+    let popup_link = ca_item.querySelector('a.artwork-image');
```
Freso: "Back in my day" `let` was not a thing. I’ll read up on `let` and be sure to not use `var` in the future. :p

mb_1200px_caa.user.js
```diff
@@ -21,22 +21,22 @@ var ca_page = document.querySelector('div#content');
...
-    if (link_list_a[i].textContent == '500px') {
-      var _500px_link = link_list_a[i];
-      break;
+    /* Add a “1200px” link to the “All sizes” list */
+    // Until https://tickets.metabrainz.org/browse/CAA-88 is resolved.
+    let link_list = ca_item.querySelector('p.small');
+    let link_list_a = link_list.getElementsByTagName('a');
+    for (let i = 0; i < link_list_a.length; i++) {
+        if (link_list_a[i].textContent == '500px') {
+            var _500px_link = link_list_a[i];
```
Freso: Why not `s/var/let/` here?

Freso: @Freso Because `let` is block scoped, so `_500px_link` would not be available outside of the `if`, let alone outside of the `for`.

murdos:
Thanks @zas

PR #238, opened 2019-11-02
Grab the upc from Bandcamp when importing.

djl:
It would also be nice to check the "This release does not have a barcode" box if a UPC isn't found but I couldn't think of a simple way to do it. Maybe someone smarter than me can figure it out :smile:


zas:
I pushed a commit (https://github.com/murdos/musicbrainz-userscripts/commit/f5869e7c95eacf708f914427f3a3063db853ede7) doing this, I didn't see this PR :(

Issue #154, opened 2017-09-08
Buttons do not show on Beatport


m0nkey-br4in:
Import script version 2017.06.13.0 in Tampermonkey v4.3.6
Chrome on Windows version 61.0.3163.79 (Official Build) beta (64-bit)

Beatport release page lacks "import to MusicBrainz" buttons (and I don't remember what they were supposed to look like)
The page looks like this:
![screenshot 2017-09-08 17 50 09](https://user-images.githubusercontent.com/8735883/30218660-47e295be-94c2-11e7-89ed-3ff43c03bb2a.png)

Tampermonkey popup shows that the script is enabled
There is no error in the JS console that looks like `"ERROR: Execution of script 'Import Beatport releases to MusicBrainz' failed! "`
There is, however, this error:
```
Uncaught TypeError: Cannot read property 'length' of undefined
    at Function.each (eval at E_c (:3:221), <anonymous>:6:2728)
    at retrieveReleaseInfo (eval at E_c (:3:221), <anonymous>:620:5)
    at HTMLDocument.eval (eval at E_c (:3:221), <anonymous>:576:17)
    at j (eval at E_c (:3:221), <anonymous>:6:26925)
    at Object.fireWith [as resolveWith] (eval at E_c (:3:221), <anonymous>:6:27738)
    at Function.ready (eval at E_c (:3:221), <anonymous>:6:29532)
    at I (eval at E_c (:3:221), <anonymous>:6:29723)
    at <anonymous>:2:398
    at b (<anonymous>:2:243)
    at t (eval at exec_fn (:1:108), <anonymous>:47:115)
```
Clicking on a link on the right that leads to its source reveals an error somewhere inside the included jQuery lib in this piece of code:
```javascript
n.extend({
    ... snip ...
    each: function(a, b, c) {
                            var d, e = 0, f = a.length, g = s(a);
                            if (c) {
                                if (g) {
                                    for (; f > e; e++)
                                        if (d = b.apply(a[e], c),
                                        d === !1)
                                            break
                                } else
                                    for (e in a)
                                        if (d = b.apply(a[e], c),
                                        d === !1)
                                            break
                            } else if (g) {
                                for (; f > e; e++)
                                    if (d = b.call(a[e], e, a[e]),
                                    d === !1)
                                        break
                            } else
                                for (e in a)
                                    if (d = b.call(a[e], e, a[e]),
                                    d === !1)
                                        break;
                            return a
                        },
    ... snip ...
```
[...]

Issue #62, opened 2015-06-17
BeatPort Pro: buttons not showing until one reloads the page


zas:
When coming from main beatport pro page, the freshly displayed release doesn't have buttons.
Workaround: just reload the page.

VxJasonxV:
#24 :)

This is specific to Greasemonkey. The script is going to have to have a dramatically different implementation in order for this to not be the case :(.

PR #244, opened 2019-12-20
Reload Playables if empty

zas:
It seems one has to be logged in to beatport for this patch to work

It should fix https://github.com/murdos/musicbrainz-userscripts/issues/154


augustjanse:
Does this also fix #62?

PR #260, opened 2020-03-29
replaces for with reduce for counting tracks

GuilhermeHideki:
Uses [const](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/const), [arrow functions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions), [reduce](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reduce) and [object destructuring](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#Object_destructuring) features, which are supported by modern browsers

### DIFF
--- a/lib/mbimport.js
+++ b/lib/mbimport.js
@@ -373,10 +373,7 @@ var MBImport = (function() {
     function searchParams(release) {
         let params = [];
 
-        let totaltracks = 0;
-        for (let i = 0; i < release.discs.length; i++) {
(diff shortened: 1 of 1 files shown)


murdos:
Thanks @GuilhermeHideki !

PR #262, opened 2020-03-31
mb_relationship_shortcuts: Fix width of subheaders

kellnerd:
This adapts the width of all subheader rows in the release table on release group pages. Usually there is a single `th`-Element with a `colspan` of 7, which now has to span the additional *Relationships* column that is inserted by this script.
In case the server code changes, only the `colspan` of the first `th` cell is incremented.

**Before:**
mb_relationship_shortcuts user js before

**After:**
mb_relationship_shortcuts user js after


murdos:
Thanks @kellnerd !

Issue #158, opened 2017-10-05
discogs importer TypeError: Cannot read property 'format' of undefined


ghlupe2:
https://www.discogs.com/de/Bruno-Walter-New-York-Philharmonic-Orchestra-Columbia-Symphony-Orchestra-Mahler-Bruno-Walter-conduct/release/9693581

V-Toll:
Same here: https://www.discogs.com/Various-Trojan-50-Box-Set/release/12282213

PR #271, opened 2020-04-22
Fix use of undefined release.discs[discindex]

antlarr:
Opening https://www.discogs.com/Ryuichi-Sakamoto-Year-Book-1985-1989/release/11652104
gives a problem because undefined (release.discs[discindex]) doesn't have
a format member in line 721
`release.discs[discindex].format.match(/(Vinyl|Cassette)/) ...`

This fixes the problem (it leaves the medium empty and stores its tracks
in the previous medium, but at least it doesn't break the script).

### DIFF
--- a/discogs_importer.user.js
+++ b/discogs_importer.user.js
@@ -703,7 +703,7 @@ function parseDiscogsRelease(data) {
 
         // Create release if needed
         let discindex = releaseNumber - 1;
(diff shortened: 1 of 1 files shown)


antlarr:
This fixes #158

Freso (approved):

PR #272, opened 2020-04-23
Add 'recording' mbid param to release seeder

loujine:
Adding the Parameter "mediums.x.track.y.recording" defined in https://musicbrainz.org/doc/Development/Release_Editor_Seeding

Not sure if this should be documented since it's an optional parameter. I just needed it for https://github.com/loujine/musicbrainz-scripts/blob/master/mb-edit-create_release_from_recording.user.js where I knew the preexisting recording MBID

### DIFF
--- a/lib/mbimport.js
+++ b/lib/mbimport.js
@@ -223,7 +223,8 @@ var MBImport = (function () {
                     total_duration += duration_ms;
                 }
                 appendParameter(parameters, `mediums.${i}.track.${j}.length`, tracklength);
-
+                appendParameter(parameters, `mediums.${i}.track.${j}.recording`,
+                    track.recording ? track.recording : '');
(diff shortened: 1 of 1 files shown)


lib/mbimport.js
```diff
@@ -223,7 +223,8 @@ var MBImport = (function () {
                     total_duration += duration_ms;
                 }
                 appendParameter(parameters, `mediums.${i}.track.${j}.length`, tracklength);
-
+                appendParameter(parameters, `mediums.${i}.track.${j}.recording`,
+                    track.recording ? track.recording : '');
```
murdos, line 227: The `appendParameter` function already checks if the input is not empty:
```suggestion
                appendParameter(parameters, `mediums.${i}.track.${j}.recording`, track.recording);
```

murdos:
Thanks @loujine !

PR #277, opened 2020-06-08
bandcamp_importer: seed barcode if present

atj:
Minor change to set the release barcode if a UPC is present.


djl:
Just FYI the UPC on the Bandcamp page is not guaranteed to be the UPC for the digital release. This was previously implemented then reverted in 6e1f053bc364568ddd26094df54e6ae31ff4c51d.

Issue #138, opened 2017-02-19
Discogs importer does not handle grouped tracks properly


eharris:
An example is the last four tracks of this Discogs entry:

https://www.discogs.com/Empire-Brass-Mozart-For-Brass/release/1996829

zas:
Is there actually 18 tracks ? If you have the CD can you provide the discid ?

I ask that because sometimes "grouped tracks" are splitted in tracklist but not on the CD, and since MusicBrainz doesn't support subtracks yet, titles (and durations) are joined to match the discid.

The main problem here is to distinguish the two cases from discogs listing. Can someone very used to discogs provides links to releases covering the two cases ?

eharris:
Definitely 18 tracks:

`e90f0612 18 150 20075 34508 41238 55323 63048 77148 100010 118963 138223 150115 167110 188578 196733 208633 234195 261535 270988 3848`

eharris:
I can only assume the tracks are marked as grouped in the discogs entry in order to avoid duplicating the piece title in each track.  So far I've found several other similar entries with these grouped tracks, and in every case I've seen so far they have been real separate tracks.

In any case, even if it turns out that there isn't a fool-proof way to determine for sure that the grouped tracks are always really separate tracks, it seems a lot safer to err on the side of treating them as separate tracks, since it would be much easier to delete potentially extra tracks than to have to go manually create them and manually enter all the details (the track name, artist, etc).

zas:
I agree, this doesn't work at all: the release which was motivating the move appears to be in the same case, it has 12 tracks per disk, but importer join subtracks, creating only 4 tracks.

References:
https://github.com/murdos/musicbrainz-userscripts/commit/45c0216ea060ea2dde3b821cb54e9428619869d4
https://www.discogs.com/release/5880212
https://musicbrainz.org/release/cd7ee0c3-ae2b-4d5e-b81d-12f7ac96185e/discids

https://github.com/murdos/musicbrainz-userscripts/commit/adf6c4fdca93e9f3cca647ebbbb7600310868091
https://www.discogs.com/release/1156598

I'll try to fix this mess as soon i have time for it if no one makes a PR for that before.

jesus2099:
Same problem here with tracks 2-02a, 2-02b, 2-25a and 2-25b.
https://musicbrainz.org/edit/54204039

mll0:
[...]

PR #281, opened 2020-06-18
discogs_importer: subtracks in Discogs to separete tracks in Musicbrainz

hirokobayashi:
Hi
I modified Discogs importer to generate separate tracks for Discogs subtracks instead of merging them to one track in musicbrainz.
The track format will be "trackname: subtrackname" which would match the classical music guideline.


jesus2099:
Hello @hirokobayashi,
But MusicBrainz is supposed to represent the CD faithfully, without adding fictitious tracks, isn't it?

hirokobayashi:
> Hello @hirokobayashi,
> But MusicBrainz is supposed to represent the CD faithfully, without adding fictitious tracks, isn't it?

This commit fixes #138

Please check the discogs document about track and heading here.
https://support.discogs.com/hc/en-us/articles/360005055373-Database-Guidelines-12-Tracklisting#Index_Tracks_And_Headings

Original discogs importer uses a heading as a track and subsequent tracks are merged into the heading.

jesus2099:
Oh thanks, in fact you already found a doc about the distinction between indeed tracks (that are 1 CD track) and headings (CD track set, logical group of several CD tracks).

I would like to test your patch with the two cases and come back here afterwards. ☺️

ke4roh:
I've run into the bug this fixes.  Here's a vote for merge!

kaysond:
Please merge this!

kaysond:
Pinging @jesus2099 @murdos @Schweinepriester @kellnerd 

I'm happy to rebase this if someone will commit to merging it in the next week.

kaysond:
I've fixed this for classical albums in a really hacked together fork here: https://github.com/kaysond/musicbrainz-userscripts/blob/discogs-classical/discogs_importer.user.js

jesus2099:
Wow!

It seems better!

| Discogs release | Current importer | @hirokobayashi + @kaysond version |
| - | - | - |
| [2 CD](https://www.discogs.com/release/16345092) | 🟢 | 🟢 |
| [Single track with indexes](https://www.discogs.com/release/350350) | 🔴 | 🟢 |

(I will update this table with more tests.

kaysond:
@jesus2099  - correct. This PR is based on a very old discogs import script. Mine was based on the latest one, but its not very good code!

jesus2099:
I don't know why and what, but there have been more recent PR merged to this script, since then.
This PR seemed like a progress.
Is there a volunteer to [re-benchmark](https://github.com/murdos/musicbrainz-userscripts/pull/281#issuecomment-1134703159) this PR vs current version?
It's not supposed to fix all Discogs tracklist mistakes, but to turn some 🔴 red lights into 🟢 green lights, like it already did, back in **2020** and then **2022**!

And then we ask some admins to merge this quickly, to avoid wasting more time benchmarking later again.

jesus2099:
@hirokobayashi could you check out this possible typo: 

https://github.com/hirokobayashi/musicbrainz-userscripts/blob/291b654de031b66a2611073a4f862c8d8cfe6f96/discogs_importer.user.js#L716

Shouldn't it be `track_new.duration` instead of `track_new.duraction`?
In which case, please fix, and this PR will be even better! :)

hirokobayashi:
@jesus2099 Thank you for review. I fixed the typo.

arsinclair:
I was going to look at it last week, but didn't find the time. Hopefully next weekend, then.

I was wondering if you can think about some ways to create automated tests for this, since with so many different ways to write things on Discogs our logic might break any time. Ideally, some way to snapshot all Discogs pages in question and run the script on top of the snapshot.

Maybe not to implement it yet, not within this PR, but just to brainstorm how it can be done.

hirokobayashi:
The function "parseDiscogsRelease" in "discogs_importer.user.js" parse a json from Discogs and generate a json for Musicbrainz.
So you can test it has regression,  as long as if you keep the input json and expected output.

I made a script to test it. But I don't know where to put it in this repo. So I'll attach it as a file.
How to use the script
``` bash
# get a Discogs track file as json
curl https://api.discogs.com/releases/1996829 > trackset.json
# generate expected result
node test.js trackset.json > trackset-expected.json
# compare the latest result with expected result
node test.js trackset.json trackset-expected.json
```
I expect test.js is in the same directory of discogs_importer.user.js

I add a mod to export the "parseDiscogsRelease" function to use from test file.

[test.js](https://github.com/user-attachments/files/28641930/test.js)

arsinclair:
Thank you, this is a great idea!

I've created some kind of testing framework based on your suggestion in https://github.com/murdos/musicbrainz-userscripts/pull/933.

Could you please take a look at it, and then ideally integrate them into your branch and update relevant fixtures that your patch fixes.

arsinclair:
I've been looking today again at the changes that this PR brings and I'm not sure this approach is entirely correct. In some situations splitting makes sense: classical music, symphonies, concertos, etc.

Good examples from the table above:
- https://www.discogs.com/release/1996829-Empire-Brass-Mozart-For-Brass
- https://www.discogs.com/release/15313328-Beethoven-Bernard-Haitink-London-Symphony-Orchestra-Symphonies-Nos-1-9
- https://www.discogs.com/release/5880212-JS-Bach-Christophe-Rousset-The-Academy-Of-Ancient-Music-Christopher-Hogwood-The-Complete-Harpsichord
- https://www.discogs.com/release/5880212-JS-Bach-Christophe-Rousset-The-Academy-Of-Ancient-Music-Christopher-Hogwood-The-Complete-Harpsichord

In other situations such division doesn't make sense. Bad examples from the table:
- https://www.discogs.com/release/350350-Ammer-Einheit-Crashing-Aeroplanes — this CD has one continuous track, confirmed via attached discid. The track names are arbitrary and it is not clear where one ends and the other begins. If we separate, we are losing track durations as well as the ability to match this release correctly.
- https://www.discogs.com/release/1156598-King-Crimson-Lizard tracks Bc1, Bc2 and Bc3 are logical division, but not actual division. On the CD they come as one track.
- https://www.discogs.com/release/8661943-Barry-Manilow-Live - a medley recording. A single track with no concrete dividers between the songs included in the medley. Similar to the first bad example, if we separate these, we're losing track durations.

This commit illustrates the changes: https://github.com/murdos/musicbrainz-userscripts/pull/281/commits/15391034ec20896e49ea2ecc99588ab2de55438c

It seems to me that MB is just lacking the plumbing to make such logical grouping possible. And I think always splitting tracks based on the logical grouping in Discogs is wrong. Maybe it makes sense to split only when track durations are actually available on discogs.

@jesus2099 @hirokobayashi and others, what do you think?

jesus2099:
Let's focus on releases, one by one, because I'm not sure what you mean.

This looks like:

> In other situations such division doesn't make sense. Bad examples from the table:
> 
> * https://www.discogs.com/release/350350-Ammer-Einheit-Crashing-Aeroplanes — this CD has one continuous track, confirmed via attached discid. The track names are arbitrary and it is not clear where one ends and the other begins. If we separate, we are losing track durations as well as the ability to match this release correctly.

| Discogs release | Current importer | @hirokobayashi + @kaysond version |
| - | - | - |
| [2020-05-10](https://community.metabrainz.org/t/merging-releases-with-different-number-of-tracks/476464?u=jesus2099): [single split track](https://www.discogs.com/release/350350-Ammer-Einheit-Crashing-Aeroplanes) | 🔴 | 🟢 |

For me it's correct, it imports a release with only one track, with multiple titles:

> **1.** _Crashing Aeroplanes: Hindenburg At Lakehurst / Voice Recorder / Value Jet 592 At Miami / To The Sky / Value Jet 592 At Everglades / Case Of Emergency / EIAI 1862 At Amsterdam / Up There / JAL 123 At Mount Osutaka / Autopilot / PanAm 103 At Lockerbie / Last Words / Baggage Claim_

The script does not separate, and this is what's intended, no?

arsinclair:
> For me it's correct, it imports a release with only one track, with multiple titles:

Current stable version imports them as one track. But the version from this PR separates them. See https://github.com/murdos/musicbrainz-userscripts/commit/15391034ec20896e49ea2ecc99588ab2de55438c

In your original table you indicated with a green circle that separating them is a desired beavior.

jesus2099:
Strange, I have installed the version from this PR (from _**Files changed tab**_ => View File => Raw), though.

But I think I have issues with Violentmonkey new MV3 version, at the moment.

hirokobayashi:
The Discogs tracklisting guidelines define how subtracks that belong to the
same physical track should be numbered:

> Sub tracks, for example, DJ mixes that comprise one track on a CD:
> Separate songs or tunes that are rolled into one track on a CD, LP, etc.,
> should be listed using a point and then a number: 1, 2, 3.1, 3.2, 3.3, 4, …
> Letters can also be used, with or without a point: A3.a, A3.b or A3a, A3b, …

Based on this rule, I added a regular-expression check to identify standard
Discogs subtrack positions. When a position matches one of these patterns,
the existing subtrack-handling routine is used instead of creating a new
track.

However, the King Crimson release uses non-standard position values such as
`Bc1` and `Bc2`. These values do not follow the standard subtrack position
formats described in the Discogs guidelines, so they cannot be handled
reliably by the position-based check alone.

PR #286, opened 2020-07-08
mb_ui_enhancements: Add a-tisket links next to Deezer, Spotify, iTune…

djl:
…s, Apple Music links.

I've been running this as a standalone userscript for a few weeks and thought someone else might like this.


murdos:
Thanks @djl !

Issue #193, opened 2018-11-12
Define licence for scripts without @licence OR define default licence in this project


murdos:

murdos:
For my part I don't any strong preference between these ones:
-  [x] X11
-  [x] Apache 2
-  [x] GPL 3

Freso:
I'd prefer GPL FWIW, but as long as it's [open source](https://opensource.org/osd-annotated) I'm happy.

zas:
> I'd prefer GPL FWIW, but as long as it's [open source](https://opensource.org/osd-annotated) I'm happy.

Same.

PR #289, opened 2020-07-21
Add Boomkat importer.

djl:
This is a first attempt at a Boomkat importer. The code isn't pretty but it does seem to pass the linter clean. It correctly detects the selected format (e.g. Digital, CD, LP, etc.) and sets the link types and media accordingly.

Reviews welcome :)

### DIFF
--- a/boomkat_importer.user.js
+++ b/boomkat_importer.user.js
@@ -1,0 +1,163 @@
+// ==UserScript==
+// @name        Import Boomkat releases to Musicbrainz
+// @description Add a button on Boomkat release pages to open MusicBrainz release editor with pre-filled data for the selected release
+// @version     2020.07.21.1
+// @license     X11
+// @namespace   https://github.com/murdos/musicbrainz-userscripts
+// @include     https://boomkat.com/products/*
+// @require     https://ajax.googleapis.com/ajax/libs/jquery/2.1.4/jquery.min.js
+// @require     lib/mbimport.js
+// @require     lib/logger.js
+// @require     lib/mbimportstyle.js
+
+// ==/UserScript==
+
+// prevent JQuery conflicts, see http://wiki.greasespot.net/@grant
+this.$ = this.jQuery = jQuery.noConflict(true);
+
+if (!unsafeWindow) unsafeWindow = window;
+
+$(document).ready(function () {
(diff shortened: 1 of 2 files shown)


atj:
Looks OK based on a brief review. The only thing I found was a duplicate require statement on line 8.

Was the `animate` parameter to `insertLink` copied from another script? The reason I ask is I've found that `slideDown()` doesn't work for me and I have to replace it with `show()`.

djl:
> Was the `animate` parameter to `insertLink` copied from another script? The reason I ask is I've found that `slideDown()` doesn't work for me and I have to replace it with `show()`.

No. The `animate` parameter is only there so the animation happens once when the page loads. Each time the user clicks on a format (e.g. "MP3" or "CD") the import/search links need to be removed, then added again with new parameters.

> The reason I ask is I've found that `slideDown()` doesn't work for me and I have to replace it with `show()`.

Interesting. Aside from the new parameter, the `insertLink` function is copied directly from the Bandcamp importer. `slideDown()` has always worked for me in both scripts.

djl:
Fixed the duplicate import.

boomkat_importer.user.js
```diff
@@ -0,0 +1,162 @@
...
+
+// prevent JQuery conflicts, see http://wiki.greasespot.net/@grant
+this.$ = this.jQuery = jQuery.noConflict(true);
+
+if (!unsafeWindow) unsafeWindow = window;
+
+$(document).ready(function () {
+    MBImportStyle();
+
+    let release_url = window.location.href.replace('/?.*$/', '').replace(/#.*$/, '');
```
murdos, line 22: Since you never change the value of this variable, you should use `const` instead of `let`.
This is true for most usage of `let` in this script.

djl:
Changed the `let`s to `const`s where possible.

murdos:
Thanks @djl ! Last thing before I merge this PR: could you indicate here, or by adding a `// @licence ` metadata in the header, under which licence you're releasing this script?
Unfortunately, there's no global licence for this repository, see https://github.com/murdos/musicbrainz-userscripts/issues/193

djl:
Done.

I went with X11 but I'm happy to change it to whatever the global license will be if/when that happens.

PR #290, opened 2020-07-27
mb_discids_detector: Don't load requirements over HTTP.

djl:
The sha1 lib from http://pajhome.org.uk/crypt/md5/sha1.js is not available over HTTPS so we need to bundle it.

`lib/sha1.js` will fail the linter but I've done this intentionally so that it's easier to verify that the bundled version is actually the one from the `http://pajhome.org.uk/crypt/md5/sha1.js`.


djl:
IIRC MusicBrainz does something odd with their disc ID calculation that might not be supported by the browser's crypto implementation. I'll try to find some time soon to look into it.

PR #300, opened 2020-09-13
Replace Lodash by built-in ES6 methods

### LINKED ISSUE #299: batch-add-recording-relationships.user.js broken since last server update
The userscript 'MusicBrainz: Batch-add “performance of” relationships'
does not work anymore since the MusicBrainz Server update 2020-09-07.

The message (when I select the "Recordings" tab) is "UncaughtReferenceError: _ is not defined"
(in this case recordings line 44 > injectedScript 248:28 which is
`var RECORDING_TITLES = _.chain($recordings)` but there are more occurences of "_.")
The console log also includes 2 times requests to "https://staticbrainz.org/MB/undefined" (answered with 404) before.

kellnerd:
Lodash has been dropped from MBS and this breaks some userscripts that rely on it.
I've replaced all occurences of `_` (Lodash) with the equivalent methods of `Array` and `Object`.

Fixes #299

### DIFF
--- a/batch-add-recording-relationships.user.js
+++ b/batch-add-recording-relationships.user.js
@@ -39,7 +39,7 @@
 // authorization.
 // ==/License==
 
-var scr = document.createElement('script');
+const scr = document.createElement('script');
 scr.textContent = `(${batch_recording_rels})();`;
 document.body.appendChild(scr);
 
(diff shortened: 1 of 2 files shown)


murdos:
Thanks @kellnerd !

PR #302, opened 2020-09-13
encylopedisque: code cleanup

murdos:


encyclopedisque_importer.user.js
```diff
@@ -1,6 +1,6 @@
 ﻿// ==UserScript==
 // @name           Import Encyclopedisque releases to MusicBrainz
-// @version        2018.2.18.1
+// @version        2020.09.13.1
```
kellnerd, line 3: Version numbers with a leading zero may cause problems

murdos: Thanks for the reminder! 😉

PR #311, opened 2020-10-06
Add hardtunes importer

basxto:
Add importer for hardtunes.com

### DIFF
--- a/hardtunes_importer.user.js
+++ b/hardtunes_importer.user.js
@@ -86,9 +86,12 @@ function retrieveReleaseInfo(release_url) {
             //duration: "0:00",
             artist_credit: []
         };
-        track.artist_credit.push(translateArtist(item.getElementsByClassName('release-list-item-artist')[0].innerText.replace(/\s/,' ')));
+        let artists = item.getElementsByClassName('release-list-item-artist')[0].childNodes;
+        // artist name and connecting strings alternate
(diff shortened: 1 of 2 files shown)


hardtunes_importer.user.js
```diff
@@ -0,0 +1,965 @@
...
+
+// Insert button into page under label information
+function insertLink(release, release_url) {
+    let edit_note = MBImport.makeEditNote(release_url, 'HardTunes.com');
+    let parameters = MBImport.buildFormParameters(release, edit_note);
+
+    document.getElementsByClassName('social-media')[0].innerHTML += MBImport.buildFormHTML(parameters) + MBImport.buildSearchButton(release);
+}
+
+// Fixed mappings for some popular labels
```
murdos, line 116: I really doubt that having an hardcoded mapping is a good idea (both for labels and artists): it's error prone and difficult to maintain.

basxto: This online shop has specialized on a few genres and name clashes within closely related genres are quite unlikely.
But I have no plans for extending that list any further, it’s only meant for more popular artists, which might be on compilations more.

murdos:
Discarding this PR, since - as previously discussed - maintaining it would be difficult since all MBIDs are hardcoded in script.

JuniorJPDJ:
@basxto you could add hardtunes artists urls to MBz and then search by URL

PR #326, opened 2020-11-19
Add support for release annotation field in mbimport

loujine:
I intend to use the annotation field in a future importer script to store release metadata from the importing website (idagio). Add support for it


augustjanse:
You should probably add it to the example object as well. `secondary_types` is missing too, for the record.

loujine:
> You should probably add it to the example object as well. `secondary_types` is missing too, for the record.

Added in https://github.com/murdos/musicbrainz-userscripts/pull/335

augustjanse:
Cool, thanks!

PR #408, opened 2021-08-05
Make the 'Edit recording comments' button appear again

kellnerd:
The script will now run after the page and all resources are loaded and page scripts have run. This will be later than before and (hopefully) after any React redraws which undid the DOM changes done by the script.

Tested and working with Violentmonkey under Firefox, for Tampermonkey document-idle is the default value anyway.

See https://community.metabrainz.org/t/541943 where the problem has also been reported and where I will ask for some additional test users with different browser / userscript manager setups.


murdos:
Thanks @kellnerd !

PR #414, opened 2021-08-21
Set recording comments: Exclude "Add release" page

kellnerd:
Script causes a type error on this page which is unintentionally matched.

### DIFF
--- a/set-recording-comments.user.js
+++ b/set-recording-comments.user.js
@@ -1,16 +1,15 @@
 // ==UserScript==
 // @name           MusicBrainz: Set recording comments for a release
 // @description    Batch set recording comments from a Release page.
-// @version        2021.8.5.1
+// @version        2021.8.22.1
 // @author         Michael Wiencek
 // @license        X11
 // @namespace      790382e7-8714-47a7-bfbd-528d0caa2333
 // @downloadURL    https://raw.githubusercontent.com/murdos/musicbrainz-userscripts/master/set-recording-comments.user.js
 // @updateURL      https://raw.githubusercontent.com/murdos/musicbrainz-userscripts/master/set-recording-comments.user.js
 // @match          *://*.musicbrainz.org/release/*
-// @match          *://*.mbsandbox.org/release/*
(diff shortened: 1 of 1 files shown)


set-recording-comments.user.js
```diff
@@ -11,6 +11,7 @@
 // @match          *://*.mbsandbox.org/release/*
 // @exclude        *musicbrainz.org/release/*/*
 // @exclude        *.mbsandbox.org/release/*/*
```
murdos, line 13: mbsandbox.org doesn't seem to be online anymore: could you remove the two mentions here and on L.11?

set-recording-comments.user.js
```diff
@@ -11,6 +11,7 @@
 // @match          *://*.mbsandbox.org/release/*
 // @exclude        *musicbrainz.org/release/*/*
```
murdos, line 12: Could you update this one so all expressions are similar? (adding prefix `*://`)

set-recording-comments.user.js
kellnerd: Sure thing, I have now removed them and made the match/exclude patterns consistent. I was not sure whether these sandboxes were internal servers or so and hesitated to remove something that did no harm.

set-recording-comments.user.js
kellnerd: Sorry, I just remembered that the `@match` patterns do not work for `@exclude` rules. You either have to use the rule as it already is or use `@exclude-match`. Please do not merge yet, I will fix it again!

kellnerd: I decided to go with `@exclude *musicbrainz.org/release/add*` which will also match possible query parameters while `@exclude-match *://*.musicbrainz.org/release/add` would have been cleaner but unfortunately incompatible with Greasemonkey.

murdos: Good catch and well done! :+1: 
Is this ready to be merged? It seems so, but I'd rather ask to be sure.

kellnerd: Yes it is 😇

PR #419, opened 2021-09-05
mb_discids_detector: Fix regex pattern for title/artists.

djl:
At least one Gazelle site recently changed from using a regular hyphen to an en dash. Update the regex to include this character.


kellnerd:
I don't use this script, but since it's a trivial change, I could probably merge it nevertheless (I'm just looking through unmerged PRs).
Do you have a link to an example where this happens?

PR #423, opened 2021-09-22
Fix "Error loading dependencies."

11xx:
Recovered the `http://pajhome.org.uk/crypt/md5/sha1.js` file from the Wayback Machine and updated the `// @require` with a gist link


djl:
(http://pajhome.org.uk/crypt/md5/sha1.js is still working for me)

But if this request is to be merged, it might be a better idea to bundle the script like I've done in [my old PR](https://github.com/murdos/musicbrainz-userscripts/pull/290) instead of hosting it in gist (which may also disappear at some point).

11xx:
oh fine if it's working, sowwy

PR #424, opened 2021-09-22
CSS for the line break in the "Import into MB" button in deezer.com

11xx:
Apparently it was an issue only noticeable in deezer.com:

Before the style
2021-09-22-08-03-53-screenshot

After the style
2021-09-22-08-05-33-screenshot


murdos:
Thanks @11xx !

PR #426, opened 2021-09-26
bandcamp_importer: Update atisket URL.

djl:
The consensus seems to be that this is the maintained instance now.


atj:
Thanks for submitting this change. I'll continue to maintain the instance (with some assistance) until a suitable replacement materialises.

PR #432, opened 2021-10-14
Show warning before edit comments on recordings used on other releases.

mattgoldspink:
The key changes are:
 - After loading the recordings for the release, make api calls for each recording's details
 - As this is happening show a message when the "Set Recording Comments" is shown that says there are still {x} recordings being loaded
 - When a recording is associated with more than one release, show inline next to the title of the recording a clear warning about how many this is associated with
 - In addition now show a warning in the "Set Recording Comments" form that some recordings need to be double checked before submitting.

Animated Preview:
image

In addition I moved the setInterval call to be after the check for the pathname, otherwise we end up
making a lot of unnecessary API requests which we won't be able to act upon.

I also changed to use the `fetch` api because `$.get` error handling is currently broken by the Sentry upgrade in MB.

### DIFF
--- a/set-recording-comments.user.js
+++ b/set-recording-comments.user.js
@@ -1,7 +1,7 @@
 // ==UserScript==
 // @name           MusicBrainz: Set recording comments for a release
 // @description    Batch set recording comments from a Release page.
-// @version        2021.8.22.1
+// @version        2021.10.14.1
 // @author         Michael Wiencek
 // @license        X11
 // @namespace      790382e7-8714-47a7-bfbd-528d0caa2333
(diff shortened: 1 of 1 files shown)


kellnerd:
You are welcome. I'm only a contributor here, just like you, I don't have the power to merge or close PRs :)
If you don't want to close this PR yourself yet, you could also wait for other opinions and let @murdos (or one of the collaborators) decide.

jesus2099:
Sorry @mattgoldspink it's [because of me](https://community.metabrainz.org/t/voting-auto-editor-request-thread/21591/892?u=jesus2099) that you created this PR. 🙇

I didn't know that @ROpdebee QoL script.

mattgoldspink:
Not a problem at all @jesus2099! @ROpdebee's script is much better and I also discovered his one to seed DJ-mix comments, which was my original goal! So it's saved me a huge amount of time :)

PR #434, opened 2021-10-17
New importer script for HDtracks

kellnerd:
The script had been [requested on the forums](https://community.metabrainz.org/t/hd-tracks-importer-script/550327?u=kellnerd) and I volunteered to implement it. In addition to the standard _Import_ and _Search_ buttons, it also has a button to submit ISRCs (via https://magicisrc.kepstin.ca). Screenshots of the running userscript and feedback can also be found in the above forum thread.

In order to make ESLint recognize `async` functions, I had to increase the ECMAScript version from 6 (= 2015) to 2017. I have also fixed `generate_README.py` which failed to decode certain Unicode characters on Windows (where the standard encoding is not UTF-8) in order to create a new entry in the README. `update_version.py` has the same problem, but a simple fix would require Python 3.10 (which finally adds an `encoding` parameter to `fileinput.input()`) and this is not worth the effort as it is easy to do this by hand for affected files.

### DIFF
--- a/hdtracks_importer.user.js
+++ b/hdtracks_importer.user.js
@@ -2,8 +2,9 @@
 // @name           Import HDtracks releases into MusicBrainz
 // @namespace      https://github.com/murdos/musicbrainz-userscripts/
 // @description    One-click importing of releases from hdtracks.com into MusicBrainz. Also allows to submit their ISRCs to MusicBrainz releases.
-// @version        2021.10.17.1
+// @version        2021.10.17.2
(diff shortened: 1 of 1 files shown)


murdos:
Great work! 
I invited you as a collaborator on this repository, so you have the rights to commit. Feel free to merge this PR whenever you think it's ready ;-)

As a side note regarding the python scripts, I was thinking about rewriting them to javascript/nodejs scripts, to only require one language stack, but I never found time to do it.

PR #439, opened 2021-11-12
vgmdb_importer: Fix websites bug

### LINKED ISSUE #404: VGMdb Importer stopped working
Up until last night the imported worked flawlessly but when I tried using it today, the import isn't showing on any release 
image
Tried reinstalling the script and changing from Tampermonkey to Violentmonkey but it isn't working on neither.
Also tried switching from Vivaldi browser to Firefox, still not working

augustjanse:
Broke when there was `websites` but no `websites['Commercial']`. Fixes #404.

### DIFF
--- a/vgmdb_importer.user.js
+++ b/vgmdb_importer.user.js
@@ -2,7 +2,7 @@
 // @name           Import VGMdb releases into MusicBrainz
 // @namespace      https://github.com/murdos/musicbrainz-userscripts/
 // @description    One-click importing of releases from vgmdb.net into MusicBrainz
-// @version        2021.7.13.1
+// @version        2021.11.12.1
 // @downloadURL    https://raw.githubusercontent.com/murdos/musicbrainz-userscripts/master/vgmdb_importer.user.js
 // @updateURL      https://raw.githubusercontent.com/murdos/musicbrainz-userscripts/master/vgmdb_importer.user.js
 // @include        /^https://vgmdb.net/(album|artist|org)/\d+
@@ -138,7 +138,7 @@ function mapUrls(vgmdbLink, stores, websites) {
         }
     }
 
-    if (websites) {
+    if (websites && websites['Commercial']) {
         for (const commercial of websites['Commercial']) {
             // Seems to fill same purpose as stores for albums
             urls.push({ url: commercial['link'] });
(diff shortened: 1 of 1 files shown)


vgmdb_importer.user.js
```diff
@@ -2,7 +2,7 @@
 // @name           Import VGMdb releases into MusicBrainz
 // @namespace      https://github.com/murdos/musicbrainz-userscripts/
 // @description    One-click importing of releases from vgmdb.net into MusicBrainz
-// @version        2021.7.13.1
+// @version        2021.11.12.1
 // @downloadURL    https://raw.githubusercontent.com/murdos/musicbrainz-userscripts/master/vgmdb_importer.user.js
 // @updateURL      https://raw.githubusercontent.com/murdos/musicbrainz-userscripts/master/vgmdb_importer.user.js
 // @include        /^https://vgmdb.net/(album|artist|org)/\d+
```
kellnerd, line 8: Thanks, the fix for the missing `Commercial` attribute itself looks good to me, but I noticed that the regex of the include rule is broken:
```suggestion
// @include        /^https://vgmdb.net/(album|artist|org)/\d+/
```
Not sure how this has been working before, since the terminating slash had been missing right from the start.
But I could swear that the userscript did also work for me (Firefox/Violentmonkey) before... Maybe I had fixed that locally and forgot to submit a PR?

kellnerd (changes requested):

vgmdb_importer.user.js
augustjanse: Huh, I guess that's right: https://wiki.greasespot.net/Include_and_exclude_rules

I don't think I every used this script with anything other than Tampermonkey. Did it stop working in Violentmonkey when you tried this patch out?

kellnerd: Yes, it did work for two of the three example releases from https://github.com/murdos/musicbrainz-userscripts/issues/404#issuecomment-927127335. Then I installed the patched version from this PR and it was no longer being injected at all, that's why I guessed that I probably had fixed my local installation.

augustjanse: Makes sense, thanks for the fix!

kellnerd (approved):

PR #440, opened 2021-11-13
Small mbimport.js fixes

jesus2099:
- Fix doc `artist_credit.mbid` (not `artist_mbid`)
- Allow no labels or no urls without crashing the script

### DIFF
--- a/lib/mbimport.js
+++ b/lib/mbimport.js
@@ -52,7 +52,7 @@
  *         {
  *             credited_name,
  *             artist_name,
- *             artist_mbid,
+ *             mbid,
  *             joinphrase
  *         },
  *         ...
--- a/lib/mbimport.js
+++ b/lib/mbimport.js
@@ -189,20 +189,24 @@ var MBImport = (function () {
         appendParameter(parameters, 'annotation', release.annotation);
 
         // Label + catnos
-        for (let i = 0; i < release.labels.length; i++) {
-            let label = release.labels[i];
-            appendParameter(parameters, `labels.${i}.name`, label.name);
-            appendParameter(parameters, `labels.${i}.mbid`, label.mbid);
-            if (label.catno != 'none') {
-                appendParameter(parameters, `labels.${i}.catalog_number`, label.catno);
+        if (Array.isArray(release.labels)) {
+            for (let i = 0; i < release.labels.length; i++) {
+                let label = release.labels[i];
+                appendParameter(parameters, `labels.${i}.name`, label.name);
+                appendParameter(parameters, `labels.${i}.mbid`, label.mbid);
+                if (label.catno != 'none') {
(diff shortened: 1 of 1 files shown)


lib/mbimport.js
```diff
@@ -189,20 +189,24 @@ var MBImport = (function () {
         appendParameter(parameters, 'annotation', release.annotation);
 
         // Label + catnos
-        for (let i = 0; i < release.labels.length; i++) {
-            let label = release.labels[i];
-            appendParameter(parameters, `labels.${i}.name`, label.name);
-            appendParameter(parameters, `labels.${i}.mbid`, label.mbid);
-            if (label.catno != 'none') {
-                appendParameter(parameters, `labels.${i}.catalog_number`, label.catno);
+        if (release.labels && Array.isArray(release.labels)) {
```
kellnerd, line 192:
```suggestion
        if (Array.isArray(release.labels)) {
```
This also works since `Array.isArray(undefined) === false`.

lib/mbimport.js
```diff
@@ -189,20 +189,24 @@ var MBImport = (function () {
...
+                }
             }
         }
 
         // URLs
-        for (let i = 0; i < release.urls.length; i++) {
-            let url = release.urls[i];
-            appendParameter(parameters, `urls.${i}.url`, url.url);
-            appendParameter(parameters, `urls.${i}.link_type`, url.link_type);
+        if (release.urls || Array.isArray(release.urls)) {
```
kellnerd, line 204: Should be AND instead of OR, but you can also use only the `isArray()` check as above:
```suggestion
        if (Array.isArray(release.urls)) {
```

lib/mbimport.js
jesus2099: Thanks @kellnerd, I added this improvement in forced pushed 137e8a55cd884a751c88d1b50ba8a32f87d77fb8. :)

lib/mbimport.js
jesus2099: Thanks @kellnerd, I never manage to use this GitHub review successfully so I added this fix in forced pushed 137e8a55cd884a751c88d1b50ba8a32f87d77fb8. :)
This `||` error was a remaining of my previous version.
I must admit I did these fixes without actually running the library for real.
I used the previous version for my script, at the moment. 
But as soon as (if) this PR is merged, I will be able to omit `labels` and `urls` from my GnuDB.org quick and dirty import script.

jesus2099:
These `artist_mbid` crash and no `labels` then no `urls` were the 3 crashes I had when trying to use the library for the first time.
The first two took me some time to understand, then the third was similar to the second.

kellnerd (approved):
Thanks, I have just tested these changes locally by removing the `labels` and `urls` from the data of an importer script and your changes prevent it from crashing (as expected).

PR #446, opened 2021-12-03
Bandcamp:  Correctly split "artist - title" & ESLint fixes

kellnerd:
Fixes ESLint errors and bumps the version (still necessary after #441 has been merged).

Correctly splits "artist - title" patterns for releases with per-track artists using a non-greedy expression for the artist as the title might contain the " - " separator (e.g. for ETI). It might be possible in certain cases that the artist credit contains the separator instead, but I'd rather accidentally have an artist as part of the title (rarer and easier to fix, especially if " - " separates multiple artists which have to be split anyway) than the artist containing parts of the title.

### DIFF
--- a/bandcamp_importer.user.js
+++ b/bandcamp_importer.user.js
@@ -114,7 +114,7 @@ const BandcampImport = {
             let title = bctrack.title;
             let artist = [];
             if (various_artists) {
-                let m = bctrack.title.match(/^(.+) - (.+)$/);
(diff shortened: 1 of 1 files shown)


kellnerd:
I would like to have a second opinion regarding the second change, since I am not entirely sure how often the two cases actually occur. Examples which I used for testing:

- Separator used for multiple artists: https://underratedrecords.bandcamp.com/album/-
- Separator as part of the title: https://manuelbackert.bandcamp.com/album/christmoss-time

PR #450, opened 2021-12-10
Refactor mb_relationship_shortcuts & Add streaming shortcuts

kellnerd:
Adds new shortcut icons for streaming and download relationships (Amazon Music, Apple Music, Bandcamp, Deezer, HDtracks, iTunes, Qobuz, Soundcloud, Spotify and Tidal so far). Very useful to tell otherwise indistinguishable releases in the same release group apart (well, their barcodes are different but...) and allows you to directly open them with the service of your choice.

I also did some refactoring to get rid of duplicated code and the icon URLs, which have been replaced by class names. MBS ships many favicons as CSS classes now, so there is no need to duplicate their data URLs here. Custom icons are still stored as URLs, except for the icons of some deprecated relationship types which I have now removed.

Additionally I have introduced two new features to make icons of ended rels less visible and to deduplicate rels by URL. This is useful for websites like Bandcamp which have e.g. purchase and stream rels and for which the same shortcut icon would have been injected twice.

See the individual commits for more details.


murdos (approved):
Great work!

PR #457, opened 2022-02-02
Set recording comments: Delay script and enable medium URL #399 work-around

jesus2099:
A very rough fix, `document-idle` dda9dcb6c14b7ac5ccafd2e1572180b7a281fd19 was eventually not enough for me, 9 times of 10, React still removes the button!

~~It does not fix #399, except if you quickly Expand all mediums, before the second elapses. ;)~~
<ins>I have added [ulugabi work-around](https://community.metabrainz.org/t/script-set-recording-comments-user-js/541943/10?u=jesus2099) for #399: Enable `/disc/123` (medium) URL.</ins>

For #399, it needs to delay the loading to the last moment, when the user clicks the button.
But I have no time for this, at the moment.

PR #459, opened 2022-02-08
recording-comments: Fill input based on recording MBID instead of index

kellnerd:
Fill the injected recording comment inputs based on the MBID of the associated recordings instead of the index. This was based on the assumption that the position of the input equals the absolute track position which is not the case if only [medium 4 of a release](https://musicbrainz.org/release/9c3f38c7-342b-4a2d-a160-6314abef8cf9/disc/4) is expanded (for example).

With the old approach the script always loads the disambiguation comments for the first tracks on the release, independent of the currently expanded medium. This lead to wrong initial comments for `/disc` URLs ("supported" since #457), see [report on the forum](https://community.metabrainz.org/t/script-set-recording-comments-user-js/541943/15?u=kellnerd).

Edit submission is already based on MBIDs, so there is no problem with that.

### DIFF
--- a/set-recording-comments.user.js
+++ b/set-recording-comments.user.js
@@ -1,7 +1,7 @@
 // ==UserScript==
 // @name           MusicBrainz: Set recording comments for a release
 // @description    Batch set recording comments from a Release page.
-// @version        2022.2.3
+// @version        2022.2.8.1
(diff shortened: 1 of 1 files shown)


jesus2099:
Yes, dkg says it works! 😁

PR #480, opened 2022-04-13
Fixing Deezer Script

### LINKED ISSUE #396: Import Deezer releases into MusicBrainz
The script does not work correctly

- the import button does not appear in most of the cases.
- The text inside the button overflows.

matteron:
Simple change to update querying for the toolbar.  Seems like they've obfuscated their class names or something, but left these `data-testid` attributes that we can use to hook into.

Fixes #396


kellnerd:
Deezer's HTML code with its scrambled class names and IDs has really become a PITA and makes custom scripts for it unnecessarily hard to maintain. While this solution seems to be working right now, it will likely break again soon.
For me (latest Firefox/Violentmonkey, haven't been using the script) the button appears only in one out of a few tries and the layout is broken, but that's most likely another issue.

I don't think that this userscript is maintained by @atj anymore since most MusicBrainz editors are now using a-tisket to import Deezer releases (of which @atj maintains a [mirror](https://atisket.pulsewidth.org.uk/)).
Maybe we should consider deprecating this userscript with a last update that informs all users of a-tisket's existence?

atj:
Whilst I must admit that no longer use the script, I think that it still has value and there are clearly people still using it.

Updating the class names doesn't seem to happen that often, so it's not a massive maintenance burden, and the fix is usually pretty trivial. Going forward I'll try and remember to submit a PR for this userscript whenever I have to update my a-tisket userscript.

arsinclair (approved):
The change looks good. The button shows up now.

murdos:
@matteron : could you update the version of the script? then I'll approve and merge this PR

arsinclair:
> @matteron : could you update the version of the script? then I'll approve and merge this PR

No reply for the last week. @murdos maybe you could update the version and push it to his fork ([it's very easy](https://stackoverflow.com/a/41588138/6184866)).

matteron:
Sorry about the long wait, but I've merged the main branch back into my repo.  Should be all good now.

arsinclair:
Next steps?

matteron:
You could just copy the script into the editor if you need it, otherwise, we're just waiting for the pr to get merged.

PR #493, opened 2022-07-12
SVG MB logo, more HTTPS + small changes

Schweinepriester:
Noticed the MB logo doesn't look so nice, e.g. on https://heilung.bandcamp.com/album/drif:

image

I've tried hard to find the logo somewhere hosted on https://musicbrainz.org/, but haven't succeeded.
@zas, maybe you know? :)

So for now the new URL is not as nice:
https://raw.githubusercontent.com/metabrainz/design-system/master/brand/logos/MusicBrainz/SVG/MusicBrainz_logo_icon.svg
But as you probably guess from the URL, it's from the official design system here: https://github.com/metabrainz/design-system/blob/master/brand/logos/MusicBrainz/SVG/MusicBrainz_logo_icon.svg

And now it looks nice:

image

Additionally did a bit of stuff around.

In case you are wondering why I removed the protocol-relative URLs:
https://www.paulirish.com/2010/the-protocol-relative-url/

> Now that SSL is [encouraged for everyone](https://www.eff.org/encrypt-the-web-report) and [doesn't have performance concerns](https://istlsfastyet.com/), **this technique is now an anti-pattern**. If the asset you need is available on SSL, then **always** use the `https://` asset.

Please test thoroughly before merging as I haven't 🙏

### DIFF
--- a/lib/mbimport.js
+++ b/lib/mbimport.js
@@ -63,7 +63,7 @@
 var MBImport = (function () {
     // --------------------------------------- publics ----------------------------------------- //
 
-    let special_artists = {
+    const special_artists = {
         various_artists: {
             name: 'Various Artists',
             mbid: '89ad4ac3-39f7-470e-963a-56509c546377',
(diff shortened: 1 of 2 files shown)


murdos (approved):
Looks fine

zas (approved):
LGTM

PR #503, opened 2022-10-11
Add script for YouTube Releases

Serene-Arc:
Thought I'd submit the userscript that I made/adapted for YouTube releases. I follow a lot of YouTube-based artists that only release their songs as singles through the platform, so this provides an easy way to import them as singles (not recordings) into the MB database.

### DIFF
--- a/youtube_release_importer.js
+++ b/youtube_release_importer.js
@@ -84,7 +84,9 @@ function mb_callback(req) {
     } else {
         div.appendChild(myform);
     }
-    waitForKeyElements('h1.title:last', add_button);
+    console.log("hit1");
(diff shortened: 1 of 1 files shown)


Freso:
Hi! Sorry for taking so long to get back to you. :) Can you please fill out the `@author` and declare a license?

Freso:
Oh, I also see you’ve adapted it from another userscript of anonymous origin. The copyright status of this is… questionable, and derivative works (such as this PR) are as well. I’d love a YouTube import script here, but it needs to be one written from scratch – or derived from a script of non-anonymous origin with a known (and compatible) license. Thanks for the PR though! ❤️

PR #518, opened 2023-01-11
deezer: wait for toolbar to appear

qubidt:
Sometimes the release information is processed faster than the page is rendered, and the buttons have nothing to attach to. Quick ~~hack~~ fix to wait for the toolbar to appear before rendering the button.

disclaimer: I don't actually know js so I don't know the best way to actually do this. Suggestions are welcome


Zipties:
Can confirm it fixed the issue for me

Issue #506, opened 2022-10-27
Musicbrainz UI enhancements ISRC bug


karpuzikov:
![image](https://user-images.githubusercontent.com/6075489/198408790-ea56c03a-540c-472d-9bf5-2100d54aae62.png)
ISRCs are broken

tigerman325:
Came here for the same reason.

texke:
Yow, made a temp workaround for this, feel free to (ab)use it :)
Replaces lines 158 - 184 with:
`    // Highlight Year in ISRCs codes
    re = new RegExp('musicbrainz.org/artist/([a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12})/recordings', 'i');
    if (window.location.href.match(re)) {
        let isrcColNo; // = ($("#content table.tbl thead th:eq(2)").text() == "Artist") ? 3 : 2;
        $('#content table.tbl thead th').each(function (index, th) {
            if ($(th).text() == 'ISRCs') {
                isrcColNo = index;
                return false;
            }
        });
        $('#content table.tbl tbody tr').each(function () {
            $(this).find(`td:eq(${isrcColNo})`).find('li').each(function() {
                let newHTML = '';
                var isrc = $(this).text();
                newHTML += `<a href='/isrc/${isrc}'><bdi><code>`;
                newHTML += `${isrc.substring(0, 5)}<b>${isrc.substring(5, 7)}</b>${isrc.substring(7)}`;
                newHTML += '</bdi></a>';
                $(this).html(newHTML);
            });
        });
    }`

Should give the desired result back (I think, I started using this script after breaking):
![image](https://user-images.githubusercontent.com/4090858/221374350-0d610194-e13b-4aae-9674-8c6a5188c340.png)

karpuzikov:
that works for me. thank you

murdos:
Fixed with #527

PR #527, opened 2023-02-25
Fix for ISRC year highlighting.

texke:
This should fix issue https://github.com/murdos/musicbrainz-userscripts/issues/506 .


murdos:
@texke : could you fix lint issues, please?

texke:
I _think_ I've done it, my development box is unavailable, so all had to happen in the Git/tampermonkey editors.

PR #537, opened 2023-06-30
beatport_importer: Add UPC and ISRC support

zyoshoka:
This PR adds registration feature of UPC and ISRC on beatport_importer.user.js by using the Beatport API and [kepstin's MagicISRC](https://magicisrc.kepstin.ca/).

### DIFF
--- a/beatport_importer.user.js
+++ b/beatport_importer.user.js
@@ -28,7 +28,7 @@ $(document).ready(function () {
         let release_url = window.location.href.replace('/?.*$/', '').replace(/#.*$/, '');
         let release = retrieveReleaseInfo(release_url, release_data);
 
-        const track_urls = release_data.tracks.map(url => url.replace('api.beatport.com', 'www.beatport.com/api')).replace(/\/$/, '');
+        const track_urls = release_data.tracks.map(url => url.replace('api.beatport.com', 'www.beatport.com/api').replace(/\/$/, ''));
         let results = [];
         const promises = track_urls.map((url, index) => $.getJSON(url).then(response => (results[index] = response)));
 
--- a/beatport_importer.user.js
+++ b/beatport_importer.user.js
@@ -28,12 +28,12 @@ $(document).ready(function () {
         let release_url = window.location.href.replace('/?.*$/', '').replace(/#.*$/, '');
         let release = retrieveReleaseInfo(release_url, release_data);
 
-        const track_urls = release_data.tracks.map(url => url.replace('api.beatport.com', 'www.beatport.com/api')).reverse();
+        const track_urls = release_data.tracks.map(url => url.replace('api.beatport.com', 'www.beatport.com/api'));
         let results = [];
         const promises = track_urls.map((url, index) => $.getJSON(url).then(response => (results[index] = response)));
 
         Promise.all(promises)
-            .then(() => results.map(result => result.isrc))
+            .then(() => results.map(({ isrc, number }) => ({ isrc, number })))
             .then(isrcs => {
                 insertLink(release, release_url, isrcs);
             });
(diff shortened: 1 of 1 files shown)


beatport_importer.user.js
```diff
@@ -24,12 +24,23 @@ if (!unsafeWindow) unsafeWindow = window;
...
     MBImportStyle();
 
-    let release_url = window.location.href.replace('/?.*$/', '').replace(/#.*$/, '');
-    let release = retrieveReleaseInfo(release_url);
-    insertLink(release, release_url);
+    $.getJSON(`https://www.beatport.com/api/v4/catalog/releases/${ProductDetail.id}`).done(release_data => {
+        let release_url = window.location.href.replace('/?.*$/', '').replace(/#.*$/, '');
+        let release = retrieveReleaseInfo(release_url, release_data);
+
+        const track_urls = release_data.tracks.map(url => url.replace('api.beatport.com', 'www.beatport.com/api')).reverse();
```
kellnerd, line 31: 1. Since you are already rewriting the API URL, you could also drop the trailing slash from the URL.
   Currently I'm seeing redirects happening for every track API call which can be avoided:
   https://www.beatport.com/api/v4/catalog/tracks/17674616/ [HTTP/2 301 Moved Permanently] -> https://www.beatport.com/api/v4/catalog/tracks/17674616 [HTTP/2 200 OK]
2. Reversing the order of the track API URLs in order to get the correct track order seems a bit unreliable to me, I would suggest to make use of the `number` attribute in each track API response instead (under the assumption that all Beatport releases only have a single medium). See my comments below.

beatport_importer.user.js
```diff
@@ -24,12 +24,23 @@ if (!unsafeWindow) unsafeWindow = window;
...
+        let results = [];
+        const promises = track_urls.map((url, index) => $.getJSON(url).then(response => (results[index] = response)));
+
+        Promise.all(promises)
+            .then(() => results.map(result => result.isrc))
```
kellnerd, line 36:
```suggestion
            .then(() => results.map({ isrc, number } => ({ isrc, number }))
```
2a. Preserve the track number for the next step.

beatport_importer.user.js
```diff
@@ -131,6 +143,16 @@ function insertLink(release, release_url) {
         )}${MBImport.buildSearchButton(release)}</li>`
     ).hide();
 
+    $(
+        '<form class="musicbrainz_import"><button type="submit" title="Submit ISRCs to MusicBrainz with kepstin’s MagicISRC"><span>Submit ISRCs</span></button></form>'
+    )
+        .on('click', event => {
+            const query = isrcs.map((isrc, index) => (isrc == null ? `isrc${index + 1}=` : `isrc${index + 1}=${isrc}`)).join('&');
```
kellnerd, line 150:
```suggestion
            const query = isrcs.map({ isrc, number } => (isrc == null ? `isrc${number}=` : `isrc${number}=${isrc}`)).join('&');
```
2b. Use the track number instead of the index.

kellnerd (commented):
Great work, thanks a lot for implementing these useful features!
Nevertheless I have two suggestions for minor improvements.

zyoshoka:
Thank you for your thoughtful reviews!
I have made the necessary revisions based on your suggestions.

beatport_importer.user.js
```diff
@@ -24,12 +24,23 @@ if (!unsafeWindow) unsafeWindow = window;
...
     MBImportStyle();
 
-    let release_url = window.location.href.replace('/?.*$/', '').replace(/#.*$/, '');
-    let release = retrieveReleaseInfo(release_url);
-    insertLink(release, release_url);
+    $.getJSON(`https://www.beatport.com/api/v4/catalog/releases/${ProductDetail.id}`).done(release_data => {
+        let release_url = window.location.href.replace('/?.*$/', '').replace(/#.*$/, '');
+        let release = retrieveReleaseInfo(release_url, release_data);
+
+        const track_urls = release_data.tracks.map(url => url.replace('api.beatport.com', 'www.beatport.com/api')).replace(/\/$/, '');
```
kellnerd, line 31:
```suggestion
        const track_urls = release_data.tracks.map(url => url.replace('api.beatport.com', 'www.beatport.com/api').replace(/\/$/, ''));
```

kellnerd (changes requested):
There's one misplaced parenthesis which has to be fixed, then we are good to go :)

zyoshoka:
Thank you for your feedback, I have fixed it!

PR #538, opened 2023-07-16
beatport_importer: Adapt to website changes

zyoshoka:
Beatport website has been updated recently, and the userscript no longer work at all. This PR adapts to these changes, but several problems still remain.

* The new API does not provide track numbers, so there are concerns about whether imported tracks are in the correct order.
* The API is called every 100 tracks, so for a release with more than 100 tracks like [this](https://www.beatport.com/release/a-state-of-trance-1000-celebration-mix-mixed-by-armin-van-buuren-extended-versions/3277362), import will not work properly.

### DIFF
--- a/beatport_importer.user.js
+++ b/beatport_importer.user.js
@@ -152,5 +152,6 @@ function insertLink(mbrelease, release_url, isrcs) {
     $('div[title="Collection controls"]').append(mbUI);
     $('form.musicbrainz_import').css({ display: 'inline-block', 'margin-left': '5px' });
     $('form.musicbrainz_import button').css({ width: '120px' });
+    $('form.musicbrainz_import button img').css({ display: 'inline-block' });
     mbUI.slideDown();
 }
--- a/beatport_importer.user.js
+++ b/beatport_importer.user.js
@@ -21,35 +21,36 @@ this.$ = this.jQuery = jQuery.noConflict(true);
 
 if (!unsafeWindow) unsafeWindow = window;
 
-$(document).ready(function () {
+$(document).ready(() => {
     MBImportStyle();
 
-    $.getJSON(`https://www.beatport.com/api/v4/catalog/releases/${ProductDetail.id}`).done(release_data => {
-        let release_url = window.location.href.replace('/?.*$/', '').replace(/#.*$/, '');
-        let release = retrieveReleaseInfo(release_url, release_data);
+    const release_url = window.location.href.replace('/?.*$/', '').replace(/#.*$/, '');
 
-        const track_urls = release_data.tracks.map(url => url.replace('api.beatport.com', 'www.beatport.com/api').replace(/\/$/, ''));
-        let results = [];
-        const promises = track_urls.map((url, index) => $.getJSON(url).then(response => (results[index] = response)));
+    const data = JSON.parse(document.getElementById('__NEXT_DATA__').innerHTML);
+    const release_data = data.props.pageProps.release;
 
-        Promise.all(promises)
-            .then(() => results.map(({ isrc, number }) => ({ isrc, number })))
-            .then(isrcs => {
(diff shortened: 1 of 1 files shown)


zyoshoka:
I fixed an importing issue that occurs when track count is over 100, and it is ready for review!

beatport_importer.user.js
```diff
@@ -95,65 +90,66 @@ function retrieveReleaseInfo(release_url, release_data) {
...
+            const query = isrcs.map((isrc, index) => (isrc == null ? `isrc${index + 1}=` : `isrc${index + 1}=${isrc}`)).join('&');
             event.preventDefault();
             window.open(`https://magicisrc.kepstin.ca?${query}`);
         })
         .appendTo(mbUI);
 
-    $('.interior-release-chart-content-list').append(mbUI);
+    $('div[title="Collection controls"]').append(mbUI);
     $('form.musicbrainz_import').css({ display: 'inline-block', 'margin-left': '5px' });
     $('form.musicbrainz_import button').css({ width: '120px' });
```
atj, line 154: The import button alignment was off in my browser (Brave on Linux). This may have been the case before the redesign but I don't think so:

image

In any case the following change fixed it for me:

```suggestion
    $('form.musicbrainz_import button').css({ width: '120px' });
    $('form.musicbrainz_import button img').css({ display: 'inline-block' });
```

atj (changes requested):
LGTM, thanks for fixing this! I found a minor display issue as noted above for which I have suggested a small change.

beatport_importer.user.js
zyoshoka: This issue occurred after the redesign and I was wondering what to do about it. Thank you for your suggestion!

zyoshoka:
The problem I mentioned above, which occurs when the track number is over 100, can be solved by sending a direct API request, as shown in the code below, but it requires a Bearer token. This token is apparently issued separately for each client (unlike what Apple Music does), so you need to get your own token somehow. A quick way is to go to the network tab of your browser and find the request you sent to the API and put the token you find in the header in the script, but this is not a smart way as you will have to renew the script every time the token expires. Is there a better way?
```js
$(document).ready(() => {
    MBImportStyle();

    const release_url = window.location.href.replace('/?.*$/', '').replace(/#.*$/, '');

    const data = JSON.parse(document.getElementById('__NEXT_DATA__').innerHTML);
    const release_data = data.props.pageProps.release;

    const initial_url = `https://api.beatport.com/v4/catalog/releases/${release_data.id}/tracks/?per_page=100`;
    const tracks = [];
    const accessToken = ''; // Type your access token here

    getTracks(initial_url, tracks, accessToken, tracks_data => {
        const isrcs = tracks_data.map(track => track.isrc);
        const mbrelease = retrieveReleaseInfo(release_url, release_data, tracks_data);

        setTimeout(() => insertLink(mbrelease, release_url, isrcs), 1000);
    });
});

function getTracks(url, tracks, accessToken, callback) {
    $.ajax({
        url: url,
        type: 'GET',
        beforeSend: xhr => xhr.setRequestHeader('Authorization', `Bearer ${accessToken}`),
        success: data => {
            data.results.forEach(result => tracks.push(result));

            if (data.next) {
                getTracks(data.next, tracks, callback);
            } else {
                callback(tracks);
            }
        },
    });
}
```

atj:
To be honest, I'd just merge this PR as is for now and not worry too much about the 0.01% of releases with > 100 tracks. You could then raise a new PR later if you find a reasonable way to resolve it.

However it's not my call, so let's see what @kellnerd's thoughts are...

kellnerd:
I agree with @atj, let's bump the version and release it to have a working userscript again.

zyoshoka:
Thank you @atj and @kellnerd. I will create an issue.

PR #542, opened 2023-08-18
Fix require links in bandcamp scripts

Sophist-UK:
Using Tampermonkey and copying and pasting these scripts, these scripts do not work without the full URL


zas:
I made a more generic fix for that, as it concerns all userscripts in this repo -> https://github.com/murdos/musicbrainz-userscripts/pull/581

PR #543, opened 2023-09-15
beatport_importer: support track length submission in milliseconds

zyoshoka:
Use `length_ms` property instead of `length`.


kellnerd (approved):
Thanks again!

My first though was that the change looked too simple to actually work, but I've quickly confirmed it with a test import and it is indeed enough to simply choose another track property for millisecond accuracy :open_mouth:

PR #548, opened 2023-11-09
Seed format acetate

vzell:
Add support for the newly introduced "Acetate" format when seeding new releases, see also https://tickets.metabrainz.org/browse/STYLE-2407


kellnerd (approved):
Thanks, I was about to ask for an example link for testing, but I've already found a few.
Looks good to me, the sizes are working as expected.

vzell:
Thx a lot.... danke schöööön

PR #558, opened 2024-03-28
Beatport: Display barcode under catalog number

atj:
This change adds a new barcode row to the release info section. If a barcode is set then it is linked to a-tisket:

image

If the barcode is null then "[none]" is displayed:

image


kellnerd (approved):
Thanks, this feature works as promised for the releases with which I have tested and the code LGTM as well :+1:
I will merge once you have bumped the version.

atj:
> Thanks, this feature works as promised for the releases with which I have tested and the code LGTM as well 👍 I will merge once you have bumped the version.

Thanks! I've added another commit to bump the version.

PR #559, opened 2024-03-28
Discogs: minor updates

atj:
This PR fixes a typo in a selector and adds a new selector for artists on VA master release pages.


kellnerd (approved):
Thank you, another case covered. I guess I rarely open VA release pages on Discogs as I have never missed these artist links.

PR #564, opened 2024-06-16
set-recording-comments.user.js: Fix @match

### LINKED ISSUE #563: Set recording comments not working if URL has query string
The `@include` statement in set-recording-comments.user.js prevents it from working when the release URL has a query string, for example the "?tport=xxxx" which appears when you use Picard. Or maybe it's some other user script that adds that, can't remember. :D Anyway, the script should probably work when a query string has been added.

ErikLevin:
Fixes #563.
Also change deprecated `@include` to `@match`.

### DIFF
--- a/set-recording-comments.user.js
+++ b/set-recording-comments.user.js
@@ -1,13 +1,13 @@
 // ==UserScript==
 // @name           MusicBrainz: Set recording comments for a release
 // @description    Batch set recording comments from a Release page.
-// @version        2022.2.8.1
+// @version        2024.6.18.1
 // @author         Michael Wiencek
 // @license        X11
 // @namespace      790382e7-8714-47a7-bfbd-528d0caa2333
 // @downloadURL    https://raw.githubusercontent.com/murdos/musicbrainz-userscripts/master/set-recording-comments.user.js
 // @updateURL      https://raw.githubusercontent.com/murdos/musicbrainz-userscripts/master/set-recording-comments.user.js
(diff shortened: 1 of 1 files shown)


set-recording-comments.user.js
```diff
@@ -7,7 +7,9 @@
 // @namespace      790382e7-8714-47a7-bfbd-528d0caa2333
 // @downloadURL    https://raw.githubusercontent.com/murdos/musicbrainz-userscripts/master/set-recording-comments.user.js
 // @updateURL      https://raw.githubusercontent.com/murdos/musicbrainz-userscripts/master/set-recording-comments.user.js
-// @include        /^https?:\/\/(\w+\.)?musicbrainz\.org\/release\/[a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12}(\/disc\/\d+|$)/
+// @match          *://*.musicbrainz.org/release/*
```
kellnerd, line 10: Unfortunately `@match` is not flexible enough to handle MB's URLs. The `release/*` wildcard matches way too many path segments which are not MBIDs and which would have to be excluded again. You've got 2 to 4 covered, but there are more, listing them all is not paracticable.
It would be simpler to keep the existing regex and extend it to allow any query parameters with `(\?.+?)?`.

ErikLevin: Alright, fixed now.

kellnerd (approved):
I have tested the new include rule and had to move the query fragment to still run on `/disc/1#cc4f12f8-5297-4fe2-9277-83bae27ef6c3` routes (where you end up when you click a track number in a tracklist).
We also could have supported optional hashes with your approach, but that could have been done before already but wasn't, so this solution might be simpler and less risky.

PR #572, opened 2024-08-29
 bandcamp_importer: add Harmony seed/import link

Aerozol:
- Added a link to seed the UPC into the Harmony importer
- Moved the seed links (Atisket + Harmony) onto their own row

### DIFF
--- a/bandcamp_importer.user.js
+++ b/bandcamp_importer.user.js
@@ -1,7 +1,7 @@
 // ==UserScript==
 // @name           Import Bandcamp releases to MusicBrainz
 // @description    Add a button on Bandcamp's album pages to open MusicBrainz release editor with pre-filled data for the selected release
-// @version        2022.5.6.1
+// @version        2024.9.6.1
 // @namespace      http://userscripts.org/users/22504
 // @downloadURL    https://raw.github.com/murdos/musicbrainz-userscripts/master/bandcamp_importer.user.js
 // @updateURL      https://raw.github.com/murdos/musicbrainz-userscripts/master/bandcamp_importer.user.js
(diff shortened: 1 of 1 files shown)


kellnerd:
_Replying to @Aerozol's [comments](https://chatlogs.metabrainz.org/libera/metabrainz/msg/5376732/) from the #metabrainz chat_

> The code works fine on my end, but those lint checks are beyond me.

It complains about some formatting issues, some of them even from before you touched this code. Don't worry, I will take care of that.

> But I was going to ask you anyway - I had two options, seeding the barcode to the initial Harmony screen, or skipping that and inserting the barcode into the URL where harmony is already doing the lookup.
> The second option is better because then it adds the Bandcamp URL’s automatically. But I have to make some decisions for the user regarding what regions to check, which has to be in the URL.

There is a fallback value for the region parameter, so we don't have to worry about that.

> Hmm I just realised that seeding the Bandcamp URL, instead of the barcode, to the initial Harmony screen would probably be the best? Anyway, enough waffling, let me know what you think!

I agree that would probably be the best option, the linked page already gives the user additional value over the script by listing the barcodes of the physical releases.
Since the Bandcamp lookup is already cached, it would still be efficient for the user to select additional providers (and a region) for a repeated lookup, IMO.
Harmony also has a query parameter `category` which can be used to avoid having to explicitly list the used providers, `category=all` would use all providers for lookup,  `category=default` only the default providers with an efficient barcode lookup.

But I am not so sure of the exact purpose of these userscript links which use other external tools to lookup releases, and what would be a sensible default (seeding the URL only I guess).
From reading #263, I think the a-tisket link was originally added to check if the barcode indeed is for the digital release rather than a physical release.
But I guess most people have since used it to import Bandcamp releases with a-tisket (by manually adding the Bandcamp link to the other seeded links).
And now we have arrived at a point where we link to Harmony which effectively acts as a (full?) replacement for the userscript 😅

It would feel sneaky for me (as the main developer of Harmony) to merge such a PR, but if it is useful for people which are using this script and @murdos and @zas are not opposed, neither am I.

murdos:
> It would feel sneaky for me (as the main developer of Harmony) to merge such a PR, but if it is useful for people which are using this script and @murdos and @zas are not opposed, neither am I.

No problem for me, if it's useful to users, let's do it.

zas:
No problem for me either ;)

Aerozol:
Thanks for looking at this in such detail kellnerd!

Not that many Bandcamp releases have barcodes, so the link definitely won't replace the Bandcamp importer!
But, as you have surmised, when there is a barcode I use the link to check if the release is on other platforms, and add all of them if so. And other tools (Atisket, Harmony) make that easier. If we can seed the BC url to Harmony I will no longer have to manually add the Bandcamp link and details afterwards!

Let me know if there's anything I can still help with, I didn't mean to dish you up a bunch of extra work :D

kellnerd:
Oh sorry, I still haven't looked into this PR again, will do it tomorrow.
I don't think there is anything left you can help with, unless you want to try updating the Harmony URL to use the BC URL instead of the barcode, should be something like `encodeUriComponent(release.url)`.

Aerozol:
Oh, no rush at all! I have this change on my personal script anyway :D
I think doing the BC URL will take you 5 min and me two hours so I'm happy to bow out! If that ends up with the Bandcamp URL's being added I will be a happy chappy, but even just the barcode link is very useful for me right now. So if it's a headache feel free to leave it for another PR/day.

kellnerd (approved):
I did the necessary changes to seed the Bandcamp URL and make the linter happy.
Ready to be released!

Issue #569, opened 2024-08-23
Deezer importer: Issues


MaxEtMoritz:
Hi, Thank you for your work in building all these import scripts.

Today while using the Deezer importer i came across two issues:
1. The importer only imports the first 25 tracks of an album, not the complete one. Probably a pagination issue?
2. Only the main artist is imported, additional artists are missing.

I looked at what Deezer returns from the API, and to resolve 1., a check if `nb_tracks` is greater than the length of the `tracks` array needs to be done. if this is the case, the next tracks need to be requested from `https://api.deezer.com/album/<album_id>/tracks?index=<track array length>` (by the way, this track list also lists track ISRCs that could be submitted to MB).

For 2., not even the dedicated track list includes all artists (unless there's a parameter that can be set to include those. Don't have a deezer account that would be necessary to check the official api docs). Requesting each individual track would include those artists.

Unfortunately i do not have the time to assemble a PR that resolves 1. since i'm going on vacation tomorrow.

PR #574, opened 2024-09-09
feat(deezer_importer): fix multi-disc imports and add ISRC submissions

MaxEtMoritz:
Partial fix for #569 (all tracks and not just top 25 tracks are imported now. Featuring artists are still missing)
Additional improvements:
- Multi-Disk releases are now correctly submitted as multiple media to MB
- ISRC submission from Deezer is now possible


MaxEtMoritz:
Currently i more-or-less copy-pasted most of the ISRC stuff from another importer script (beatport). But in the long run i guess it's better to make it a library function (`MBImport.buildIsrcButton(release)` or similar). What do the devs think about this?

When i have time i could try to do that if this is desired.

MaxEtMoritz:
Another small improvement to ISRC submission:
since MagicISRC now also supports submitting and seeding an edit note, i added the default edit note for ISRC submissions.
@murdos could you have a look at the PR?

murdos:
@MaxEtMoritz : could you fix the lint issues?

MaxEtMoritz:
@murdos sorry, done now.
Switched PC since i last worked on it and forgot to `npm install`, so the git hooks didn't trigger...
And i wasn't aware i had set my VSCode formatter to Prettier and not ESLint, so only half of the linting was done.

murdos (approved):

murdos:
@MaxEtMoritz : no problem. And sorry for not having reviewed your PR earlier 😅

jesus2099:
Tested with **multi-disc 25+ tracks** https://www.deezer.com/album/978557941 which could not be imported with previous version.
**ISRC submission** is also working good with that same example.

arsinclair:
Not sure why this was never merged. @MaxEtMoritz thank you for your contribution!

arsinclair (approved):

PR #575, opened 2024-09-10
boomkat_importer: Fix for recent Boomkat changes.

djl:
Rather than fight with jQuery conflicts, just use regular JS.

### DIFF
--- a/boomkat_importer.user.js
+++ b/boomkat_importer.user.js
@@ -10,22 +10,24 @@
 // @require     lib/mbimportstyle.js
 // ==/UserScript==
 
-async function onLoad() {
+function onLoad() {
     MBImportStyle();
 
     const release_url = window.location.href.replace('/?.*$/', '').replace(/#.*$/, '');
-    let release = await retrieveReleaseInfo(release_url);
+    let release = retrieveReleaseInfo(release_url);
     insertLink(release, release_url, true);
 
     // Update the release info when a different tab/format is
     // selected. We need a timeout here due to a slight delay between
     // clicking the tab and the tab's classes (and the tracklist)
     // being updated.
-    document.querySelector('ul.tabs li.tab-title a').addEventListener('click', function () {
-        setTimeout(function () {
-            release = retrieveReleaseInfo(release_url);
-            updateLink(release, release_url, false);
-        }, 150);
+    document.querySelectorAll('ul.tabs li.tab-title a').forEach(function (elem) {
+        elem.addEventListener('click', function () {
+            setTimeout(function () {
+                release = retrieveReleaseInfo();
+                updateLink(release, release, false);
+            }, 500);
+        });
     });
 }
 
@@ -58,7 +60,7 @@ function getLinkType(media) {
     return MBImport.URL_TYPES.purchase_for_mail_order;
 }
 
-async function retrieveReleaseInfo(release_url) {
+function retrieveReleaseInfo(release_url) {
     const releaseDateStr = document.querySelector('span.release-date-placeholder').textContent.replace('Release date: ', '');
     const releaseDate = new Date(releaseDateStr);
     const artist = document.querySelector('h1.detail--artists').textContent.trim();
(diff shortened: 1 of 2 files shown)


kellnerd (commented):
This is just a quick review as I have never used the Boomkat importer, but the changes seem straightforward, so I am just dropping a few comments for now.

boomkat_importer.user.js
```diff
@@ -1,43 +1,36 @@
...
-    let release = retrieveReleaseInfo(release_url);
+    let release = await retrieveReleaseInfo(release_url);
     insertLink(release, release_url, true);
 
     // Update the release info when a different tab/format is
     // selected. We need a timeout here due to a slight delay between
     // clicking the tab and the tab's classes (and the tracklist)
     // being updated.
-    $('ul.tabs li.tab-title a').click(function () {
+    document.querySelector('ul.tabs li.tab-title a').addEventListener('click', function () {
```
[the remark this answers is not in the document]
djl: Good spot. Fixing led me down a rabbit hole. Turns out the tracklist on the page is now actually usable so we don't need to resort to extra HTTP requests and HTML parsing!

boomkat_importer.user.js
```diff
@@ -108,56 +99,53 @@ function retrieveReleaseInfo(release_url) {
...
+    let tracks = [];
+    const response = await fetch(tracklistURL);
+    const body = await response.text();
+    const doc = document.implementation.createHTMLDocument('');
+    doc.body.innerHTML = body;
+    doc.querySelectorAll('div.table-row.track.mp3.clearfix a').forEach(function (link) {
+        tracks.push({
+            artist_credit: MBImport.makeArtistCredits([link.dataset.artist]),
+            title: link.dataset.name,
+            duration: link.dataset.duration,
```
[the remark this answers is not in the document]
djl: Pretty much. I'm not even sure if `data-duration` was there before, but it's there now so we should use it.

boomkat_importer.user.js
```diff
@@ -28,16 +21,18 @@ $(document).ready(function () {
...
-    $('ul.tabs li.tab-title a').click(function () {
-        setTimeout(function () {
-            release = retrieveReleaseInfo(release_url);
-            updateLink(release, release_url, false);
-        }, 150);
+    document.querySelectorAll('ul.tabs li.tab-title a').forEach(function (elem) {
+        elem.addEventListener('click', function () {
+            setTimeout(function () {
+                release = retrieveReleaseInfo();
+                updateLink(release, release, false);
```
kellnerd, line 28:
```suggestion
                updateLink(release, release_url);
```
This function never supported a third parameter, it was already wrong before this PR. Since the animate parameter is also gone for `insertLink` now, it should also be removed from all calls of that function. (I can't comment directly as you haven't touched these lines.)

boomkat_importer.user.js
```diff
@@ -1,23 +1,16 @@
 // ==UserScript==
 // @name        Import Boomkat releases to Musicbrainz
 // @description Add a button on Boomkat release pages to open MusicBrainz release editor with pre-filled data for the selected release
-// @version     2020.07.21.1
+// @version     2024.09.10.1
```
kellnerd, line 4: When I wanted to install this userscript for testing, I realized that it is the only script for which the README has no install button. If you add the following two metadata lines, `tools/generate_README.py` will add one:

```suggestion
// @version     2024.09.10.1
// @downloadURL    https://raw.githubusercontent.com/murdos/musicbrainz-userscripts/master/boomkat_importer.user.js
// @updateURL      https://raw.githubusercontent.com/murdos/musicbrainz-userscripts/master/boomkat_importer.user.js
```

kellnerd (commented):
I just realized that you are also the original author of this script, so whenever you are happy with the changes, I can merge them.

While I only did two quick tests with the old and the new version of the script [using this random release](https://boomkat.com/products/making-records-home-recordings-c-1890-1920), I noticed that the new version seeds an empty tracklist if I have not switched between the release variant tabs before. Once I have done that, the tracklist is populated.
So maybe a longer initial delay is needed or the dynamic loading of the tracklist is still required for that case?
(Tested with Firefox 130.0 and Violentmonkey 2.20.0)

djl:
> I just realized that you are also the original author of this script, so whenever you are happy with the changes, I can merge them.
> 
> While I only did two quick tests with the old and the new version of the script [using this random release](https://boomkat.com/products/making-records-home-recordings-c-1890-1920), I noticed that the new version seeds an empty tracklist if I have not switched between the release variant tabs before. Once I have done that, the tracklist is populated. So maybe a longer initial delay is needed or the dynamic loading of the tracklist is still required for that case? (Tested with Firefox 130.0 and Violentmonkey 2.20.0)

Just spotted the same thing. It still happens occasionally even with the delay. I'll probably go back to fetching the tracks manually again if I can't figure it out.

djl:
Should be good to merge now. Quick summary of the changes:

- Added updateURL/downloadURL.
- Went back to manually fetching/parsing the tracklist.
- Moved the import/search buttons to just below the album title. This was necessary to prevent breaking the layout of the page when a release had a lot of media types.

boomkat_importer.user.js
```diff
@@ -60,7 +57,12 @@ function getLinkType(media) {
     return MBImport.URL_TYPES.purchase_for_mail_order;
 }
 
-function retrieveReleaseInfo(release_url) {
+async function fetchTracksAndInsertLink(release_url, element) {
```
kellnerd, line 60: Please give `element` a more descriptive name like `formatElement` (or `format_element`), otherwise it is hard to follow the purpose of this value across multiple function calls.

kellnerd (commented):
Sorry for the long delay, I was travelling and this PR slipped off my radar.
One minor comment before I will merge 😇

boomkat_importer.user.js
djl: Good point. Fixed.

kellnerd (approved):

PR #578, opened 2024-10-03
bandcamp_importer: Harmony link auto-checks other services

Aerozol:
Makes the Harmony link auto-search/check the other available platforms in Harmony, saving 5 clicks every time.

I'm not sure why anybody clicking this link would not want those fields checked, if they are only interested in BC data they can use this Bandcamp importer script to import.

### DIFF
--- a/bandcamp_importer.user.js
+++ b/bandcamp_importer.user.js
@@ -408,7 +408,7 @@ $(document).ready(function () {
         document.querySelector('div #trackInfoInner').insertAdjacentHTML(
             'beforeend',
             `<div id="mbimport_upc" style="margin-bottom: 2em; font-size: smaller;">UPC: ${upc}<br/>
-            Import: <a href="https://harmony.pulsewidth.org.uk/release?url=${encodeURIComponent(release.url)}">Harmony</a>
+            Import: <a href="https://harmony.pulsewidth.org.uk/release?url=${encodeURIComponent(release.url)}&deezer=&itunes=&spotify=&tidal=&beatport">Harmony</a>
             | <a href="https://atisket.pulsewidth.org.uk/?upc=${upc}">a-tisket</a></div>`
         );
     }


Aerozol:
Hey @kellnerd, what do you think about this change? Works for me locally. This is quite a important change for my workflow, saves a LOT of clicks.

A couple of things I am not sure about:
- I haven't entered any country presets into the URL, but it seems to work fine without
- It would be useful to have a URL platform selection code like "ALL", so this link doesn't have to be updated everytime a new platform is added to Harmony (just saying :P)

kellnerd:
Took me a while to get back to this, but that is exactly the thing which I tried to explain in your last PR #572. So obviously I failed 😅

Having said that, I still want Harmony to remember the user's choice for preferred providers and regions and avoid pushing external tools towards specifying defaults which not every user likes.
I think cookies could solve this issue without needing additional user interaction (like clicking a button), but I haven't tried to implement that yet.

Aerozol:
Ohhhhh, sorry! So:
`Import: <a href="https://harmony.pulsewidth.org.uk/release?url=${encodeURIComponent(release.url)}&category=all">Harmony</a>`

Your explanation was good but I was already dealing with a lot of new information and I obviously didn't absorb everything! Sorry to have to pick up the discussion again.

I understand re. not wanting to push defaults, but the current defaults (nothing checked) make the link useless - it is quicker to just copy and paste the barcode into Harmony. I'm sure there is an edge case where someone does not want to lookup/submit all the streaming links, but I can't imagine what it would be!

I will update the PR to use `&category=all` if you agree that's better than no providers being checked, otherwise this can be closed.

Aerozol:
Thanks again @kellnerd, I've made the change.

p.s. can't wait to try the new Harmony features!! I imagine you are also in the same boat, still catching up on post-summit life :-0

kellnerd (approved):

Issue #579, opened 2024-10-06
[discogs] ReferenceError: Can't find variable: MBLinks


jmariani:
Running in Safari using userscripts extension

akritm:
> Same for me in Chrome tampermonkey. And I see errors in console:
> 
> ```
>  @require: couldn't load @require from URL 'lib/mbimport.js': internal error
>  @require: couldn't load @require from URL 'lib/logger.js': internal error
>  @require: couldn't load @require from URL 'lib/mblinks.js': internal error
>  @require: couldn't load @require from URL 'lib/mbimportstyle.js': internal error
> chrome-extension://d…5d8-9caa96c2eb3d:49 Uncaught (in promise) ReferenceError: MBLinks is not defined
>     at Proxy.<anonymous> (userscript.html?name…-9caa96c2eb3d:49:17)
>     at Pt (<anonymous>:10:89)
>     at userscript.html?name…8-9caa96c2eb3d:1:88
>     at window.__f__m2c9s3xc.rco (userscript.html?name…-9caa96c2eb3d:1:355)
>     at Pt (<anonymous>:10:89)
>     at r.setMessageListener.r (<anonymous>:86:186)
>     at <anonymous>:89:107
>     at _ (<anonymous>:22:406)
> ```
> 
> Seems like external libs stopped working for some reason in userscript manager. `@require` instruction just don't load them.

fyi - temporary solved it changing to direct paths to github in tampermonkey script e.g.:
```
// @require        https://raw.githubusercontent.com/murdos/musicbrainz-userscripts/refs/heads/master/lib/mbimport.js
// @require        https://raw.githubusercontent.com/murdos/musicbrainz-userscripts/refs/heads/master/lib/logger.js
// @require        https://raw.githubusercontent.com/murdos/musicbrainz-userscripts/refs/heads/master/lib/mblinks.js
// @require        https://raw.githubusercontent.com/murdos/musicbrainz-userscripts/refs/heads/master/lib/mbimportstyle.js
```

zas:
I got this issue with Bandcamp Importer and TamperMonkey 5.3.1 (deployed on 2024-10-16), relative links ceased to work (with similar errors on console).

zas:
See https://github.com/murdos/musicbrainz-userscripts/pull/581

kellnerd:
Upstream issue report: https://github.com/Tampermonkey/tampermonkey/issues/2201

zas:
[...]

Issue #524, opened 2023-01-25
Bandcamp importer does not show up at all (Firefox 109.0, Violentmonkey + FireMonkey)


Viktini:
Title explains it. I've tested it on a bandcamp.com page, with Violentmonkey and FireMonkey with Firefox 109.0, and the Bandcamp importer doesn't show up. It's crossed out with a red line on Violentmonkey if it's used with joks' [Show publish date](https://greasyfork.org/en/scripts/420662-bandcamp-show-publish-date/code) userscript.

There's nothing in the F12 console that could be pointed towards the script.

kellnerd:
Which version of Violentmonkey are you using? There was a bug in a recent version that broke many userscripts in Firefox, but this should be fixed with VM 2.13.7 and later.
I'm currently using the latest beta 2.13.8.4 (should be equivalent to 2.13.9) and the Bandcamp importer appears for me.

Viktini:
I'll try the VM beta now, thanks for letting me know!

kellnerd:
Probably not needed, the (non-beta) 2.13.9 has already been released on AMO. Were you using an older version like 2.13.5? If not, you may have encountered a different issue.

Viktini:
Nope, latest version on both the release and the beta.

Byter09:
Can confirm. I don't see any button. Just installed a userscript for the first time using violentmonkey and there's no difference on the album page I'm testing with.

Versions:
```
Violentmonkey: 2.13.9
Firefox: 109.0
```
Album I tried it with:
https://karlcasey.bandcamp.com/album/white-bat-xx

New to this, but if there's anything I can do to help out, let me know!

Byter09:
It shows 1.
![grafik](https://user-images.githubusercontent.com/16060697/216277060-f6228494-99cb-4fef-8af8-226aea62ed87.png)

And I don't have any other addons that touch bandcamp or musicbrainz. Is there a log of some kind I can check to see what the script is doing? :o

kellnerd:
It might be that the other addon looks totally unrelated, you won't know whether another addon is the culprit unless you try to disable it.

The script should log only one line in the browser console (Ctrl+Shift+K) if it was successful: `[INFO] Parsed release: ...`

Byter09:
Oh I didn't even think about checking the console. I get this:
```
[Violentmonkey][Import Bandcamp releases to MusicBrainz] SyntaxError: expected expression, got '<'
[...]

PR #581, opened 2024-10-17
Use absolute URLs in @require statements

zas:
- relative links do not always work (we had multiple reports about this)

See https://github.com/murdos/musicbrainz-userscripts/pull/542 and https://github.com/murdos/musicbrainz-userscripts/issues/579

### DIFF
--- a/bandcamp_importer.user.js
+++ b/bandcamp_importer.user.js
@@ -1,7 +1,7 @@
 // ==UserScript==
 // @name           Import Bandcamp releases to MusicBrainz
 // @description    Add a button on Bandcamp's album pages to open MusicBrainz release editor with pre-filled data for the selected release
-// @version        2024.10.17.1
+// @version        2024.10.17.2
 // @namespace      http://userscripts.org/users/22504
 // @downloadURL    https://raw.github.com/murdos/musicbrainz-userscripts/master/bandcamp_importer.user.js
 // @updateURL      https://raw.github.com/murdos/musicbrainz-userscripts/master/bandcamp_importer.user.js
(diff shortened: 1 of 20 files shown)


kellnerd (commented):
Hmm, using absolute URLs breaks installations from branches other than `master` and for forks, so I am generally not a fan of that.
I think such issues are bugs of the userscript extensions which should be fixed there, on the other hand we got multiple such reports for different userscript extensions already (#524 has been fixed in VM), so it may be worth it 🤷

P.S. Sorry for the merge conflict, I can fix that once/if we decide to merge this PR.

zas:
> Hmm, using absolute URLs breaks installations from branches other than `master` and for forks, so I am generally not a fan of that.

Yes,  I agree with that too.
But we had issues in the past and they continue, and scripts break everytime :(

Not really happy with this PR either :( but not sure what to do.

bandcamp_importer.user.js
```diff
@@ -1,17 +1,17 @@
 // ==UserScript==
 // @name           Import Bandcamp releases to MusicBrainz
 // @description    Add a button on Bandcamp's album pages to open MusicBrainz release editor with pre-filled data for the selected release
-// @version        2024.9.6.1
+// @version        2024.10.17.1
```
kellnerd, line 4:
```suggestion
// @version        2024.10.17.2
```
I just noticed that this version has already been released today.

murdos:
Issue will be solved by packing userscripts with rollup. See https://github.com/murdos/musicbrainz-userscripts/pull/709

PR #611, opened 2025-04-21
Fix deps

zas:


.github/workflows/checks.yml
```diff
@@ -13,7 +13,7 @@ jobs:
             - name: 'Install dependencies'
               run: npm install
             - name: 'Run linter'
-              run: npm run lint
+              run: ESLINT_USE_FLAT_CONFIG=false npm run lint
```
murdos, line 16: Thanks. I'll try to migrate the configuration file in another PR, unless you plan to do it.

murdos (approved):

.github/workflows/checks.yml
zas: It would be great if you could do it. Thanks.

Issue #482, opened 2022-04-19
Qobuz importer button does not show


Annis3110:
I've tried on both the qobuz store and the qobuz web player but couldn't find the button on either of them. I've logged in to both my qobuz account (free account with no subscription) and musicbrainz account.
![IMG_20220419_202316_431](https://user-images.githubusercontent.com/76511040/164033134-ca2355f8-6a1b-453c-8635-a65df4770b65.png)
![IMG_20220419_202448_011](https://user-images.githubusercontent.com/76511040/164033149-c8e4730e-1ba7-443f-9964-dad5cdc9422a.png)
I'm using Violetmonkey's and edge's latest version. Please tell me if you need the version number of violetmonkey extension and edge.

sleepuntilspring:
im having the same issue as well

redactedscribe:
Still broken. Easiest way I've found for the time-being to get the bare minimum into MusicBrainz is the highlight and copy the tracklisting in full, put it into a text editor like VS Code, then use Alt + Click (or whichever other method) to define cursors for each track, and then batch sort the lines into the MBz track parser format:

```
1. Title - Artist (00:00:00)
2. Title - Artist (00:00:00)
...
```

Then just paste it into the track parser on the MBz Add Release page and fill the rest.

Some links about the broken script:

https://community.metabrainz.org/t/qobuz-importer-script/569614
https://community.metabrainz.org/t/qobuz-script/585934

Serene-Arc:
So I did a bit of poking and the reason that this script is failing is quite simple and rather inescapable. Qobuz has decided to unpublish their API and seems to be phasing it out. The API call that the script waits for never comes and there's no documentation or way to make a call ourselves.

From what I can tell, the only way forwards for this userscript vis-a-vis Qobuz is to parse the page directly, which will lead to less and lower-quality information. Unfortunately there's not currently a way around that.

realbaylen:
:(

DoktorSultan:
> Easiest way I've found

MP3Tag has an export function, in which you can define how to export the tages to a plain text file.
With this you can simply create a txt export compatible with the track parser for any downloaded files, if there is no (working) importer available.

Just discovered I have the same issue: Qobuz importer doesn't do anything...

Serene-Arc:
[...]

PR #624, opened 2025-05-11
DO NOT MERGE. Qobuz fixes and page source scraping

HenryGessau:
This PR is not meant for merging into the main branch.

This is an experiment to see/show what data is available in the page source for Qobuz albums, if we were to abandon use of the undocumented API.

Using the undocumented Qobuz API is tricky and unreliable because the API auth key/token must be intercepted or scraped, and Qobuz can change it at any time. Unfortunately, the API provides some key data that does not seem to be in the page source, like the label catalog number, the Qobuz release event date, and ISRCs.

I tried to provide comments about the data that I did find in the page source and how it is structured.

Nevertheless, this patch contains some useful fixes, which might be enough for #482:
- correctly placing the MB search/submit buttons
- finding the cover art 

☝️ @jeffbyrnes

(I included type hints for `lib/mbimports.js`, although these are really just for my own use to help my IDE help me keep track of all the data elements.)

### DIFF
--- a/lib/mbimport.js
+++ b/lib/mbimport.js
@@ -2,7 +2,7 @@
 //                                          MusicBrainz Import helper functions
 ///////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 
-/*
+/**
  * How to use this module?
  *
  * - First build a release object (see expected format below) that you'll fill in from source of data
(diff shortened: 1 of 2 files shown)


HenryGessau:
See #628 instead.

PR #633, opened 2025-06-01
feat(bandcamp_importer): Bandcamp importer should use batching when searching for MusicBrainz releases

arsinclair:
- just stumbled upon the Jira thread that release relationship lookup supports batching now;
- implemented a new function to batch process multiple URLs for MusicBrainz links;
- adjusted the existing logic to utilize the new batch processing feature for discography pages on Bandcamp.


murdos:
Good idea :)

PR #642, opened 2025-06-10
discogs barcode null check

### LINKED ISSUE #641: Discogs barcodes aren't null-checked, only field existence is checked
On ["Beyond Pt. 2" by Organon](https://www.discogs.com/release/2253359-Organon-Beyond-Pt2) the Barcode field exists but is empty. 
Causes `TypeError: identifier.value is null` at line 652: https://github.com/murdos/musicbrainz-userscripts/blob/a6e17fd64ce498cd7d1d3efcdce7e1b3d461f7a4/discogs_importer.user.js#L647-L656

Firepal:
UNTESTED!

should fix #641

### DIFF
--- a/discogs_importer.user.js
+++ b/discogs_importer.user.js
@@ -648,7 +648,7 @@ function parseDiscogsRelease(discogsRelease) {
     // Barcode
     if (discogsRelease.identifiers) {
         $.each(discogsRelease.identifiers, function (index, identifier) {
-            if (identifier.type === 'Barcode') {
+            if (identifier.type === 'Barcode' && identifier.value !== null) {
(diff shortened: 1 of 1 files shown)


jesus2099:
_UNTESTED!_?
Haven't you have tested this against the #641 example, at least?

Firepal:
@jesus2099 Gosh I forgot about this, well I think I'd have tested it since then!

I think I made this patch through GitHub really hastily, for some reason. Maybe I didn't realize I could literally test the script right in my browser? It's weird, I did make personal userscripts at some point. Whatever.

Finally(?!) tested using the Organon release, it does parse the release and shows the links! Tada. (There's a minified React error, which I assume is Discogs testing in production...)

PR #679, opened 2025-08-22
feat(discogs_importer): speed up lookups by using batching

arsinclair:


arsinclair:
I encountered a weird bug with closures, I'll have to troubleshoot it later. For now, I'm converting the PR to Draft.

PR #694, opened 2025-09-28
feat(qobuz_importer): add Release page linking with MB

arsinclair:
image

- also fix button styles, as right now their positioning is quite off;
- didn't really touch Classical music logic, since I never work with it and don't know the importing rules;
- release pages seem to support locale-independent URLs, but artist pages and label pages do not: Qobuz returns 404 if locale or any other parts of the slug are removed. So we will have to keep it as is and look up accordingly.

For reference:
1. Release URL: https://www.qobuz.com/us-en/album/fujiwarp-01-hiromichi-fujiwara/vulp1p7pxc6vc
2. Artist URL: https://www.qobuz.com/us-en/interpreter/hiromichi-fujiwara/22742844
3. Label URL: https://www.qobuz.com/us-en/interpreter/hiromichi-fujiwara/22742844

### DIFF
--- a/qobuz_importer.user.js
+++ b/qobuz_importer.user.js
@@ -1,7 +1,7 @@
 // ==UserScript==
 // @name         Import Qobuz releases to MusicBrainz
 // @description  Add a button on Qobuz's album pages to open MusicBrainz release editor with pre-filled data for the selected release
-// @version      2019.03.26.0
+// @version      2025.09.28
 // @namespace    https://github.com/murdos/musicbrainz-userscripts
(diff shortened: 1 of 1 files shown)


arsinclair:
Hey @murdos! Can we have some traction here, do you have any comments maybe?

murdos:
@arsinclair : I'm trusting you, thanks for your work.
FYI I'm not contributing to MusicBrainz since a couple of years, so I don't use any userscripts. I'm just in minimal maintenance mode here.

arsinclair:
Thank you very much!

PR #735, opened 2025-10-27
chore: use renovate instead of dependabot

murdos:
Renovate handles mono-repos and automerge out of the box


kellnerd:
Can we please configure this bot to be less noisy? I've received at least a few dozen (useless) emails from new automated PRs within the last few minutes...

PR #742, opened 2025-11-13
fix: Use more secure matching, and exlude edit pages

jesus2099:
Will also avoid ongoing MBS jQuery conflict:
https://tickets.metabrainz.org/browse/MBS-14188
https://github.com/metabrainz/musicbrainz-server/pull/3667

Spotted by @chaban-mb in:
https://github.com/jesus2099/konami-command/issues/938


arsinclair (approved):
Elegant!

PR #774, opened 2026-01-25
fix: fixed 25 track limit on Deezer import

mancube:
Deezer imports capped out at 25 tracks, which breaks releases that have more than 25 tracks, this fixes that problem using the API for fetching all track data, correctly importing into MusicBrainz.


jesus2099:
This duplicates the more complete and earlier #574.

arsinclair:
@mancube, thank you for taking a stab at this. An earlier PR which addresses this issue had just been merged, so I have to close this one, as at a glance it seems to not add anything else.

Feel free to open another PR if you find more issues with the script. Thanks!

PR #780, opened 2026-01-30
discogs: fix missing import/mbUI actions

### LINKED ISSUE #777: Discogs importer is now missing from Discogs release pages.
The button to import from Discogs has disappeared off of the release pages. I still see all the links to MB from there, but the import is gone.

jaawerth:
Resolves #777

There was an update to the Discogs UI such that the reference element with selector #release-marketplace no longer appears when the user is logged in, though it does when logged out.

To make the fallbacks more maintainable, I adjusted it to cycle through them in a loop instead of using if-else blocks. Let me know if it's preferable the other way, but I thought this made it more readable/easier to maintain when adjusting and annotating fallsbacks.

As for adjustments to the fallbacks:
- Left #release-marketplace as-is. The fact that it's still there when logged out suggests this may be a temporary state of affairs.
- Added #shopping-box-host as the first fallback, as this section contains the same elements #release-marketplace did before.
- Removed the broken-since-2021 fallbacks #statistics, div.section.collections, and div.section.social
- Added #release-stats, which is present on the page and points to the stats section previously referenced by the #statistics

### DIFF
--- a/discogs_importer.user.js
+++ b/discogs_importer.user.js
@@ -2,7 +2,7 @@
 
 // @name         Import Discogs releases to MusicBrainz
 // @description  Add a button to import Discogs releases to MusicBrainz and add links to matching MusicBrainz entities for various Discogs entities (artist,release,master,label)
-// @version      2025.08.13
+// @version      2026.01.29
 // @namespace    http://userscripts.org/users/22504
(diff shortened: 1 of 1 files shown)


murdos:
Thanks @jaawerth !

PR #782, opened 2026-01-31
Update bandcamp_importer.user.js

uniquePWD:
Working version. Feel free to improve.


arsinclair:
Thank you, but the current version is also working. 

I can see again, that you're using an AI coding agent to completely rewrite the script. Although there's no harm in using an AI agent to **solve a specific problem**, be prepared that such submissions will undergo an additional scrutiny.

What problem are you trying to solve with the script?

uniquePWD:
The issue was previously reported by me. Out of 100 album loads, the script would work like once. Perhaps it's because I'm using Firefox for Android, but this comprehensively fixes it.

PR #884, opened 2026-05-28
feat(beatport_importer): switch from atisket import to Harmony

### LINKED ISSUE #873: Beatport: switch to Harmony over atisket
[Harmony](https://harmony.pulsewidth.org.uk/) natively supports beatport releases, so it makes sense to use it over atisket, which generally seems to be on its way out.

I'd open a PR but I don't know TypeScript very well (if at all) so I don't know how to test the script locally from the .ts source
But I believe updating [`insertMBButtons`](https://github.com/murdos/musicbrainz-userscripts/blob/master/src/userscripts/beatport_importer/index.ts#L198) should be pretty simple
```
const harmonyURL = 'https://harmony.pulsewidth.org.uk/release?category=preferred';
const releaseID = release_url.match(/release\/[^/]+\/(\d+)/)?.[1];
let spanHTML = ``;
if (!mbrelease.barcode) {
    spanHTML = `<a href="${harmonyURL}&url=${release_url}">[none]</a>`;
}
else {
    spanHTML = `<a href="${harmonyURL}&beatport=${releaseID}&gtin=${mbrelease.barcode}">${mbrelease.barcode}</a>`;
}
```

- I'm not sure why `encodeURIComponent()` was being used on a value that's always a number, so I've dropped its usage in my example, but it can be reimplemented if necessary.
- `releaseID` is also exposed in [`getBeatportReleaseData.ts`](https://github.com/murdos/musicbrainz-userscripts/blob/master/src/userscripts/beatport_importer/utils/getBeatportReleaseData.ts) in multiple spots: [parsed from the URL](https://github.com/murdos/musicbrainz-userscripts/blob/master/src/userscripts/beatport_importer/utils/getBeatportReleaseData.ts#L155) or better yet, from the parsed JSON as [`initialReleaseID`](https://github.com/murdos/musicbrainz-userscripts/blob/master/src/userscripts/beatport_importer/utils/getBeatportReleaseData.ts#L172C15-L172C31), but again, I don't know how to work with this TypeScript structure 💀 so getting it from `release_url` seemed like the easier choice to me.
- The current ternary operator to assign `spanHTML` becomes hard to read if we also use a URL when the barcode is [none], so a simple `if...else` seems cleaner to me

arsinclair:
- if a barcode is not found for some reason (its type is nullable), we will still try to match the release using the URL since Harmony seems to support that.
- also port the code to the old JS lib, as I'm planning to reuse the button in the bandcamp script as well.
- a clickable barcode approach was extremely hard to discover in the current version, so I'm opting for a button instead.
- for now to avoid crowding, I put the button right next to the barcode, but in the future I'm thinking to move the entire buttons array to its own new row.

image

Closes #873

### DIFF
--- a/lib/mbimport.js
+++ b/lib/mbimport.js
@@ -12,6 +12,8 @@
  *     var formHtml = MBImport.buildFormHTML(parameters);
  * - Addinionally, you can inject a search link to verify that the release is not already known by MusicBrainz:
  *     var linkHtml = MBImport.buildSearchLink(parsedRelease);
+ * - Or a Harmony import link (pass barcode when available, otherwise release_url):
+ *     var harmonyHtml = MBImport.buildHarmonyButton({ barcode: parsedRelease.barcode, release_url: 'https://...' });
  *
  * Expected format of release object:
  *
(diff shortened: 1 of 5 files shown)


arsinclair:
@kellnerd could you please take a look at this PR and check that the way I'm building Harmony URLs makes sense. Danke!

arsinclair:
I've tested it and it works fine for me so I think this can be merged, but if we ever need to change how Harmony URLs are generated, I'm happy to do it.

lib/mbimport.js
```diff
@@ -121,6 +123,56 @@ const MBImport = (function () {
         return html;
     }
 
+    // compute HTML of Harmony import link
+    function fnBuildHarmonyButton({ barcode, release_url }) {
+        const harmonyWithBarcodeURL = barcode
+            ? `https://harmony.pulsewidth.org.uk/release?gtin=${barcode}&category=all`
+            : `https://harmony.pulsewidth.org.uk/release?url=${encodeURIComponent(release_url)}&category=musicbrainz`;
```
kellnerd, line 130: See https://harmony.pulsewidth.org.uk/about for the supported lookup parameters.
 
1. There is no category "musicbrainz", if you want to enforce a MB lookup by barcode you have to add `&musicbrainz`. But since Harmony already checks that the URL is not linked to any MB release, I wouldn't recommend this. It is usually just an additional request which finds nothing, and the userscript already provides a button to search for the release on MB for the extra cautious users.
2. In my opinion `category=all` is just a waste of resources which I should probably remove now that multiple more useful categories have been added. (For example it is highly unlikely that there is a significant overlap between Beatport and a physical release DB like Discogs or one of the Korean providers.) My recommendation would be to always use `category=preferred` to leave the user in control.
3. Even if the barcode is known, it is still beneficial to also add the URL to the lookup, for some providers URL lookups are faster and/or less ambiguous than barcode lookups.

arsinclair: > 1. There is no category "musicbrainz", if you want to enforce a MB lookup by barcode you have to add &musicbrainz. But since Harmony already checks that the URL is not linked to any MB release, I wouldn't recommend this.

I understand that there's a blind spot here which is why I added this parameter: there can be a release on MB with a barcode (imported from another provider) which is not linked to the current provider. Given how temperamental Beatport recently, this is actually often the case: beatport lookups by barcode are not working for the last half a year and we often have releases in MB with all supported providers + GTIN but without the Beatport relationship. With the `musicbrainz` parameter it allows to match those releases much earlier, without waiting for the MB Import Form to indicate that there's a duplicate already.

> In my opinion category=all is just a waste of resources which I should probably remove now that multiple more useful categories have been added.

I agree, the idea here was to not hardcode specific providers in the URL, so that when the new ones are added we don't have to remember to update the script.

Perfect, I didn't know about `category=preferred`, I'll switch to it.

> Even if the barcode is known, it is still beneficial to also add the URL to the lookup, for some providers URL lookups are faster and/or less ambiguous than barcode lookups.

Good, I didn't know this is supported.

arsinclair: I made some changes in https://github.com/murdos/musicbrainz-userscripts/pull/900

PR #887, opened 2026-05-28
feat(set-recording-comments): Show contextual info in edit note

jesus2099:
- **Show userscript name** (hardcoded, as GM_info is not available with this userscript injection style)
- **Show release URL** (using same style as the MBS relationship editor contextual edit notes)
- I've also allowed the userscript to run on `musicbrainz.eu` server

Example test edit note
----------------------

https://test.musicbrainz.org/edit/100304628

Before my change:

```




testing empty heading and trailing lines




```

After my change:

```
testing empty heading and trailing lines

—
Set recording comments for a release
Entered from: https://test.musicbrainz.org/release/4867a85b-4288-396d-84d8-fc3739271ec7
```

### DIFF
--- a/set-recording-comments.user.js
+++ b/set-recording-comments.user.js
@@ -201,7 +201,7 @@ function setRecordingComments() {
             $submitButton.prop('disabled', false).text('Submit changes (marked red)');
         } else {
             let editNote = $('#recording-comments-edit-note').val().trim();
-            editNote += `\n\n—\nSet recording comments for a release\nEntered from: ${location.href}`;
+            editNote += `\n\n—\nSet recording comments for a release\nEntered from: ${location.href} '''''${document.querySelector('h1')?.textContent}'''''`;
             let makeVotable = document.getElementById('make-recording-comments-votable').checked;
 
             activeRequest = $.ajax({
--- a/set-recording-comments.user.js
+++ b/set-recording-comments.user.js
@@ -1,7 +1,7 @@
 // ==UserScript==
 // @name         MusicBrainz: Set recording comments for a release
 // @description  Batch set recording comments from a Release page.
-// @version      2026.5.29
+// @version      2026.5.31
 // @author       Michael Wiencek
 // @license      X11
 // @namespace    790382e7-8714-47a7-bfbd-528d0caa2333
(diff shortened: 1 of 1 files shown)


jesus2099:
I've also allowed the userscript to run on `musicbrainz.eu` server.

arsinclair (commented):
Thank you, it's looking good. I think it would be good to further improve the edit note in order to clarify that this is a script, and where it is coming from.

E.g. something like this:

```
Entered from https://test.musicbrainz.org/release/4867a85b-4288-396d-84d8-fc3739271ec7 using "Set recording comments for a release" from https://github.com/murdos/musicbrainz-userscripts
```

jesus2099:
Hi @arsinclair,

Before I submit new commit(s),
What do you think of this format?

---

with COOL ENTITY LINKS userscript

(with COOL ENTITY LINKS userscript)

---

without userscripts

(without userscripts)

arsinclair:
Hey! Thank you, it looks great, as soon as the script version is not hardcoded and actually reflects version information from the script metadata.

jesus2099:
I've pushed the (I think) last commit for this PR, @arsinclair. :)

Step by step 5 commits but not many changes.
You can squash, if you want.

set-recording-comments.user.js
```diff
@@ -200,7 +197,9 @@ function setRecordingComments() {
             $inputs.prop('disabled', false);
             $submitButton.prop('disabled', false).text('Submit changes (marked red)');
         } else {
-            let editNote = $('#recording-comments-edit-note').val();
+            let editNote = $('#recording-comments-edit-note').val().trim();
+            editNote += `${editNote ? '\n\n—\n' : ''}Entered from '''''${document.querySelector('h1')?.textContent}''''' at ${location.href}`;
```
arsinclair: Hey! I just have one comment here, not sure how I missed it before.

The way you're accessing release title is potentially unsafe, since it relies on the assumption that the release title is written within h1 tag and that h1 tag is the first one on the page. If MusicBrainz main site were to introduce another H1 tag before the current one (even if invisible, like in a modal window that would be hidden most of the time) it would grab incorrect text.

I think we have to be more specific here, so something like this `document.querySelector('span.release-title-line > h1')?.textContent` instead.

jesus2099: Somehow I thought a new h1 would be less likely than a parent class change.
But maybe it's better, yes, in the end.
I'm pushing a new commit. :)

arsinclair: Both are equally likely I think, but if we go with your initial approach, then relase title will be outright wrong while appearing correct; whereas with a more specifc selector it will write `'undefined'` in the edit note, which is probably better, since it draws more attention. It could also be updated to fold into an empty string.

arsinclair (approved):

jesus2099:
Thanks and see you next time. 🤟

PR #930, opened 2026-06-17
fix(bandcamp_importer): namespace artist and label cache keys on discography pages

### LINKED ISSUE #795: Bandcamp Importer causes `Invalid artist_credit` MB error when importing certain releases
Bandcamp importing error on some pages: `Invalid artist_credit.names.0.mbid`.

Example: https://hibernate.bandcamp.com/album/the-state-of-shining

Image

### LINKED ISSUE #797: Bandcamp Importer: linked labels are displayed with artist icon
Should be label icon.

chaban-mb:
_Code and PR written by AI-tool_

## Problem

Fixes #795, #797

On discography/band pages (e.g. `https://hibernate.bandcamp.com`), the script looks up both artist and label relations for the page URL. Both lookups were sharing the same bare URL as their cache key, causing two distinct collision bugs:

1. **Label MBID seeded as artist credit** — when a label was found, its MBID was cached under the bare URL. On subsequent album page visits, `resolveMBID(root_url)` (line 631) would find that cached **label** MBID and incorrectly assign it to `release.artist_credit[0].mbid`, causing MusicBrainz to reject the import with:
   > `Invalid artist_credit.names.0.mbid`

2. **Symmetrical risk** — an artist MBID cached under the bare URL could equally be picked up by a label lookup on a different page visit.

### Evidence from localStorage cache (before fix)

```json
{
    "https://hibernate.bandcamp.com": {
        "urls": ["https://musicbrainz.org/label/ebd021f8-89b9-4363-8977-3751be350572"]
    }
}
```

A **label** MBID stored under the bare URL key — read by `resolveMBID()` as if it were an artist.

## Fix

Namespace both cache keys so artist and label entries are fully disjoint:

```diff
-mblinks.searchAndDisplayMbLink(cleanURL, 'artist', insertLinkCb, undefined, onSearchComplete);
-mblinks.searchAndDisplayMbLink(cleanURL, 'label', insertLinkCb, undefined, onSearchComplete);
+mblinks.searchAndDisplayMbLink(cleanURL, 'artist', insertLinkCb, `artist:${cleanURL}`, onSearchComplete);
+mblinks.searchAndDisplayMbLink(cleanURL, 'label', insertLinkCb, `label:${cleanURL}`, onSearchComplete);
```

The `label:` prefix matches the pattern already used on album pages (line 610). The `artist:` prefix is added symmetrically (as suggested in this PR's review).

The album-page artist MBID resolution is updated to match:

```diff
-let artist_mbid = mblinks.resolveMBID(root_url);
+let artist_mbid = mblinks.resolveMBID(`artist:${root_url}`);
```

### Cache after fix

```json
{
    "artist:https://hibernate.bandcamp.com": { "urls": [] },
    "label:https://hibernate.bandcamp.com": {
        "urls": ["https://musicbrainz.org/label/ebd021f8-89b9-4363-8977-3751be350572"]
    }
}
```

Label MBID is namespaced. `resolveMBID(`artist:${root_url}`)` correctly finds nothing (empty artist result), so no bogus MBID is seeded into the artist credit.

## Testing

Tested on https://hibernate.bandcamp.com/album/the-state-of-shining (the example from #795). After clearing the localStorage cache and revisiting the page, the import button works correctly without the `Invalid artist_credit` error.

### DIFF
--- a/bandcamp_importer.user.js
+++ b/bandcamp_importer.user.js
@@ -628,7 +628,7 @@ function init() {
 
         if (release.artist_credit.length == 1) {
             // try to get artist's mbid from cache
-            let artist_mbid = mblinks.resolveMBID(root_url);
+            let artist_mbid = mblinks.resolveMBID(`artist:${root_url}`);
             if (artist_mbid) {
                 release.artist_credit[0].mbid = artist_mbid;
             }
@@ -777,7 +777,7 @@ function init() {
             }
         };
 
-        mblinks.searchAndDisplayMbLink(cleanURL, 'artist', insertLinkCb, undefined, onSearchComplete);
+        mblinks.searchAndDisplayMbLink(cleanURL, 'artist', insertLinkCb, `artist:${cleanURL}`, onSearchComplete);
         mblinks.searchAndDisplayMbLink(cleanURL, 'label', insertLinkCb, `label:${cleanURL}`, onSearchComplete);
     }
 }
--- a/bandcamp_importer.user.js
+++ b/bandcamp_importer.user.js
@@ -1,7 +1,7 @@
 // ==UserScript==
 // @name         Import Bandcamp releases to MusicBrainz
 // @description  Add a button on Bandcamp's album pages to open MusicBrainz release editor with pre-filled data for the selected release
-// @version      2026.06.07.1
+// @version      2026.06.20
 // @namespace    http://userscripts.org/users/22504
 // @downloadURL  https://raw.github.com/murdos/musicbrainz-userscripts/master/bandcamp_importer.user.js
 // @updateURL    https://raw.github.com/murdos/musicbrainz-userscripts/master/bandcamp_importer.user.js
--- a/bandcamp_importer.user.js
+++ b/bandcamp_importer.user.js
@@ -778,7 +778,7 @@ function init() {
         };
 
         mblinks.searchAndDisplayMbLink(cleanURL, 'artist', insertLinkCb, undefined, onSearchComplete);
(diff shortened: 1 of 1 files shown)


arsinclair:
I suspect that it is the same problem that leads to #797. Could you verify that this fix also resolves that problem?
Example release: https://dronarivm.bandcamp.com/album/air-signs

image

Also, do you think it makes sense to prefix artist lookups as well, so there's never an ambiguity?

chaban-mb:
> Example release: https://dronarivm.bandcamp.com/album/air-signs

Inserts a label link for me there. But this won't fix existing cache. That would require waiting for it to expire, manually clearing or bumping cache key version.

Also when a Bandcamp page is linked to both an artist and a label it will only show one of them. (So far it's always the label)
E.g. https://altersynth.bandcamp.com/


> Also, do you think it makes sense to prefix artist lookups as well, so there's never an ambiguity?

Can't hurt I guess.

arsinclair:
> Inserts a label link for me there. But this won't fix existing cache.

I guess it depends on the order it was cached. Yes, we can just bump the cache version to invalidate it, once fixed.

> Also when a Bandcamp page is linked to both an artist and a label it will only show one of them.

Maybe both icons can be displayed. I don't see why not. Whether to pre-fill the input form with both the artist and the label, that's another question.

arsinclair:
OK, so can you add the artist prefix too please?

chaban-mb:
> OK, so can you add the artist prefix too please?

Done

arsinclair (approved):

arsinclair:
Thank you for this PR. I took the liberty to also bump the localstorage cache version and push it to your branch, since I want it to get merged soon. Hope you don't mind.

PR #932, opened 2026-06-20
feat: add ampwall and subvert importers.

djl:


arsinclair:
Thank you for this PR.

Could you provide links to releases on both platforms to test the scripts. Additionally, could you address all eslint issues present?

arsinclair:
Also, would it be possible to rewrite the scripts with Typescript, since there's currently an effort underway to convert all of them to Typescript?

It would make sense to do this in order to avoid rewriting them later. But if this is too much, then no problem. We can do it later.

djl:
Two random URLs to test:

- https://ampwall.com/a/mellowharsher/album/good-grief-a-mellow-harsher-collection
- https://www.subvert.fm/the-radio-room-recording-company/the-bones-of-february

Typescript rewrite incoming later today.

djl:
Silenced the eslint issues.

Started working on the Typescript conversion but doing it Propertly™ is going to take too long right now.

arsinclair:
Thank you.

Why is it going to take a lot of time? Are any of the pieces missing?

The only shared code I haven't converted to typescript is mblinks.js I think

djl:
I was assuming that it had to be converted with full types, like the Beatport converter. If it's just a quick conversion then yeah, easy enough.

arsinclair:
I guess it definitely makes sense to have it strongly typed instead of relying on _any/unknown_ types, since this is the whole point of Typescript. But you don't have to create extensive types, I think partial types are also fine: only those that you're actually using.

If that's a blocker, then just leave it and we can address it in the future.

PR #937, opened 2026-06-21
chore: replace Prettier with Oxc

arsinclair:
- Oxc Formatter is a modern performance-optimised code formatter with a full feature parity with Prettier, so it's mostly a drop-in replacement.


arsinclair:
The checks are failing on this PR, this is expected: I wanted to separate the swap of the formatter from the formatting changes in the code.

Next PR will reformat the code with the new tool.

PR #942, opened 2026-06-21
chore: introduce oxlint

arsinclair:
- migrate a few initial hardcoded rules to oxlint and address issues that eslint for some reason didn't catch;
- enable oxlint checks in the CI;
- force vscode to use tsgo and explicitly spell out oxc as a default formatter for typescript (for some reason it wasn't inferring it correctly for all files).

### DIFF
--- a/oxlint.config.ts
+++ b/oxlint.config.ts
@@ -21,9 +21,6 @@ export default defineConfig({
         es6: true,
         node: true,
     },
-    globals: {
-        unsafeWindow: 'writable',
-    },
     overrides: [
         {
             files: ['**/*.user.js'],
--- a/oxlint.config.ts
+++ b/oxlint.config.ts
@@ -1,0 +1,51 @@
+import { defineConfig } from 'oxlint';
+
+export default defineConfig({
+    plugins: ['eslint', 'typescript', 'node', 'oxc', 'import'],
+    options: {
+        typeAware: true,
+    },
+    categories: {
+        correctness: 'off',
+        suspicious: 'off',
+        style: 'off',
+        perf: 'off',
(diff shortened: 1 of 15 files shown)


oxlint.config.ts
```diff
@@ -0,0 +1,48 @@
...
+        'no-inner-declarations': 'warn',
+        'no-global-assign': 'warn',
+        'no-redeclare': 'warn',
+        'no-self-assign': 'warn',
+        // 'no-undef': 'warn', // TODO: Too many odd issues to fix, will be addressed later
+        'no-useless-concat': 'warn',
+        'no-useless-escape': 'warn',
+        'no-var': 'warn',
+    },
+    ignorePatterns: ['node_modules/**', '.git/**'],
```
murdos, line 47: Congrats for the steady work!

FYI `.git`  is already ignored (https://oxc.rs/docs/guide/usage/linter/ignore-files.html#default-ignores) and `node_modules`  too since it's included in ` .gitignore`

murdos:
Do you plan to remove eslint?
Having both might be a bit misleading.

oxlint.config.ts
arsinclair: Great find, thank you for the suggestion.

Patched in #952.

arsinclair:
> Do you plan to remove eslint?

Yes, it is very slow and brings a big dependency tree into the lockfile. I think there's no valid reason to keep using it.

Migration is not a small feat, since I want to have zero warnings with the oxlint results, so I'm going to be moving the rules incrementally. Doing it this way is also safer, as it ensures that no rules are missed in the process.

PR #943, opened 2026-06-21
tests: use vitest for discogs tests and snapshot testing

murdos:
This is a just proposal, feel free to decline it.

BTW this was co-authored by Claude.


arsinclair:
This is an awesome idea, I should have thought of it myself instead of creating a custom solution 😅

As we progressively move all scripts to TypeScript introducing automated testing will be the most logical next step.

PS. I didn't notice this PR until now and I already merged a few things that led to conflicts with your branch. I hope you don't mind me resolving them for you to reduce friction.

oxfmt.config.ts
```diff
@@ -12,7 +12,7 @@ export default defineConfig({
     endOfLine: 'lf',
     insertFinalNewline: true,
     proseWrap: 'preserve',
-    ignorePatterns: ['node_modules/**', '.git/**', 'pnpm-lock.yaml'],
+    ignorePatterns: ['node_modules/**', '.git/**', 'pnpm-lock.yaml', 'tests/**/snapshots/**'],
```
arsinclair: I don't understand this part. Why do we want to exclude them from being formatted, at the same time we're running the formatter script within `test:discogs:update-fixtures`? That will just run the formatter over the rest of the codebase, while explicitly skipping the snapshots directory.

Maybe instead it would be better to run the formatter [explicitly over the the snapshots and fixtures directories](https://oxc.rs/docs/guide/usage/formatter/cli.html#available-positional-items) after we have regenerated them, while not excluding them from the formatter script by default?

murdos:
Thanks for the feedback :)

> PS. I didn't notice this PR until now and I already merged a few things that led to conflicts with your branch. I hope you don't mind me resolving them for you to reduce friction.

Sorry I didn't notice your additional commits, and I just forced-pushed with my own resolution of conflicts.

oxfmt.config.ts
murdos: Indeed, things were a bit messy.
I clarified the situation with 41e2f5ebf4189972443b517328c18e09373bd143: fixtures are handled by the update-fixtures script and formatted.
Snapshots are fully handled by vitest, and not formatted because their format is controlled exclusively by the test runner.

arsinclair: Yes, I understood what you're doing, my question was why is done this way? What harm is there if we don't make any distinctions and the 'format' script would format everything?

murdos: My concern is that in the code of the test you have to format the result returned by the system under test in exactly the same way oxformat will format it.
It might be easy, or very complicated to replicate the formatting of oxformat. That's why if prefer the strategy: "strictly capture the snapshot of what is returned by the system under test without altering it". It's simple and without traps.

arsinclair: Wait, `toMatchFileSnapshot` compares strings? I was thinking that it does deep equality check between two objects 🤯

arsinclair: Perhaps we should use https://vitest.dev/api/expect.html#tomatchobject instead to compare the two, so sudden formatting changes don't introduce failures.

The formatting of the API response is probably not something that we need to test, since one layer above oxfmt sits JSON.stringify() which already ensures uniformity and will not fail or reorder things at random.

Issue #916, opened 2026-06-02
Release type guessing logic fails on releases with 7-8 tracks


chaban-mb:
*Written with and analyzed by AI*

The `fnGuessReleaseType` function in `mbimport.js` contains a logical gap for releases containing exactly 7 or 8 tracks.

`perhaps_EP` checks for `<= 6` tracks, while `perhaps_album` checks for `> 8` tracks. As a result, a release with 8 tracks and a duration over 30 minutes sets `perhaps_album` to `false`. 

Because of this, the condition `if (perhaps_album && duration_mn > 30)` evaluates to `false`, and the function returns an empty string `""` instead of `"album"`.

Additionally, explicit string tokens in the release title (like `"EP"` or `"Single"`) should ideally override strict duration rules.

Example release where type detection fails:
https://volumo.com/album/8721466924472-train-dreams-ep

arsinclair:
We don't have a script to support volumo.com, isn't it?

chaban-mb:
Not yet. I'm working on something:
https://github.com/chaban-mb/userscripts/blob/volumo-importer/feat/add-volumo-importer/src/Volumo%20MusicBrainz%20Importer.user.js

Based on murdos libraries but with own buttons, lookup and caching. (Actually nothing is cached persistently, just in memory)

Since I couldn't find a public API documentation I had to reverse-engineer and have to deduce exactly how the fields are filled. Most importantly I don't know yet how to best handle the different dates. 
Generally it seems quite similar to Beatport except that releases seem to often appear later instead of sooner (Beatport has many releases available in advance) 
However Volumo also has exclusives that are not available elsewhere. So it would be a valuable provider.

This release for example is an exclusive and is dated 29 May 2026 but the original date is displayed as 26 June 2026:
https://volumo.com/album/5054289829871-feel-the-beat-ep

See also:
https://community.metabrainz.org/t/stop-importing-wrong-release-dates-from-streaming-sites/814960/18?u=chaban

arsinclair:
Thank you, yes I initially failed to see the connection between userscripts here and Volumo.

It makes sense, I'll take a look at this bug.

Thank you for sharing the amazing work for the Volumo provider. If you ever decide to integrate your userscripts in this repo, you're most welcome. Just let me or Murdos know.

arsinclair:
[...]

PR #953, opened 2026-06-23
feat(common): improve Release Type detection algorithm

arsinclair:
- the changes are visible in the code;
- this is to address #916


chaban-mb:
Looks good to me.

<details><summary>But AI points out a possible bug</summary>
<p>

### The 2-track fallback bug

Take a look at the sequential logic in the `else` branch:

```typescript
} else {
    // no duration, try to guess by the number of tracks only.
    if (perhaps_EP) return "EP";
    if (perhaps_single) return "single";
    if (perhaps_album) return "album";
}

```

Because `perhaps_EP` triggers for any track count between 2 and 6, a standard 2-track release with missing durations will always hit that first `if` and return "EP".

In practice, a 2-track release—usually an A-side/B-side pair, an instrumental, or a remix bundle—is almost always a Single. Checking the EP threshold first causes immediate misclassifications here.

**The fix:** Reorder the sequence so the logic catches 1- or 2-track releases as a single before checking the EP threshold.

### A different way to handle this

For some context on how to make this even more reliable, look at how `Harmony Enhancements.user.js` handles multi-track singles. It completely skips the reliance on arbitrary time brackets or track counts.

Instead of looking at duration, it strips common technical terms from the track titles (things like *Remix*, *Instrumental*, *Edit*, *Live*, or *Version*) and compares what is left. If a multi-track release deduplicates down to a single unique title, the script overrides the status to a Single because it knows the release just contains alternate versions of the same song.

Relying strictly on track numbers and durations is always going to hit edge cases—especially with modern streaming singles that often bundle 3 or 4 tracks. Implementing a title-deduplication check like that would be a great way to make this guesser bulletproof.
 
</p>
</details>

arsinclair:
> In practice, a 2-track release—usually an A-side/B-side pair, an instrumental, or a remix bundle—is almost always a Single. Checking the EP threshold first causes immediate misclassifications here.
> The fix: Reorder the sequence so the logic catches 1- or 2-track releases as a single before checking the EP threshold.

This is often the case for Vinyl records, but not always true, especially for a bunch of electronic genres.

It we reorder the sequence, then EP branch will never trigger for releases without duration, since Single branch with 1..6 tracks will include the 2..6-track branch for EP. Additionally the set EP vs. Single sets intersection for releases with a present duration will change in ways that I cannot really see clearly. 

> Instead of looking at duration, it strips common technical terms from the track titles (things like Remix, Instrumental, Edit, Live, or Version) and compares what is left. If a multi-track release deduplicates down to a single unique title, the script overrides the status to a Single because it knows the release just contains alternate versions of the same song.

I suppose this is a great idea. I will try to add it in the next iteration. But I don't think it should completely replace the type-guesser based on the track numbers/durations, because then we will be left with quite a large number of releases which could be easily classified based on those parameters. Instead, I think we should first try to normalise song names and deduce if it's a Single, and if it yields no results then fall back on the current algo.

Let me know your thoughts.

chaban-mb:
> This is often the case for Vinyl records, but not always true, especially for a bunch of electronic genres.

I'm aware of that. I added quite a lot of techno stuff where EPs are common for 2-track releases. 

Regarding the Harmony Enhancements script, Harmony doesn't do much release type guessing instead relying on the providers, which also [classify based on track count and duration](https://community.metabrainz.org/t/remix-collections-of-one-song-single-or-ep/628484/9). 

> Instead, I think we should first try to normalise song names and deduce if it's a Single, and if it yields no results then fall back on the current algo.

Agree, such singles are easy to detect and rarely not a single per artist/label. If the release title contains EP it should still take precedence.

murdos:
BTW some unit tests could probably be added for verifying the expected behavior :)

arsinclair:
Hey, I made changes to hopefully address all remaining problems. Single heuristics also implemented and tests added.

@chaban-mb could you also review please and give it a go?

chaban-mb:
> @chaban-mb could you also review please and give it a go?

I can't give you a code review but I can test it. It works for the [example from earlier](https://github.com/murdos/musicbrainz-userscripts/issues/916#issuecomment-4623616276). I'll try to test it some  more.

murdos (approved):

arsinclair:
Thanks. I'm going to inject this change into my local userscript first and run it like that for a week/two. If you don't find any issues and I am happy with how it works, then I'll merge this.

arsinclair:
Works ok

PR #982, opened 2026-07-07
internetarchive_importer: add new importer

Jormangeud:
This importer adds releases from the Internet Archive.
Currently supporting https://archive.org/details/*

### DIFF
--- a/internetarchive_importer.user.js
+++ b/internetarchive_importer.user.js
@@ -1,0 +1,223 @@
+// ==UserScript==
+// @name         Import Internet Archive releases to Musicbrainz
+// @description  Add a button to Internet Archive pages to open MusicBrainz release editor with pre-filled data for the selected release
+// @version      2026.07.06.0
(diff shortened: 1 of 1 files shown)


Jormangeud:
Here are some collections to test with:
https://archive.org/details/audio?and%5B%5D=subject%3A%22music%22&and%5B%5D=mediatype%3A%22collection%22

Libraries are currently using the full url.
I did not add this to README yet.
The various / multiple artist stuff could be probably moved to the mbimport lib.

There is no easy way to know what is failing with the lint check without output or access to the binary.

arsinclair:
Hey Jormangeud,

The formatting changes are failing the check. You can run `pnpm format` to fix it, subsequently you can also run eslint locally to see if it passed.

Would it be possible to ask you to convert the script to TypeScript? There's an effort underway to move all scripts to TS and it would be much easier if the new script was initially written in TS.

PR #1054, opened 2026-08-29
Use release title for release page master search

jesus2099:
Current userscript searches for link title: [“See all versions”](https://musicbrainz.org/search?query=See%20all%20versions&type=release_group&indexed=1).
I made this change, so that the master link MB release group search uses current Discogs release page title, instead.

> [image](https://www.discogs.com/release/3132981-Various-The-Hits-Tape)
> The master link MB release group search, in yellow

> [IMG_20260829_155829](https://musicbrainz.org/search?query=See%20all%20versions&type=release_group&indexed=1)
> The useless “See all versions” MB search

I have ran: 

> `npx eslint discogs_importer.user.js`

With no errors, before submitting this merge request (PR).

### DIFF
--- a/discogs_importer.user.js
+++ b/discogs_importer.user.js
@@ -2,7 +2,7 @@
 
 // @name         Import Discogs releases to MusicBrainz
 // @description  Add a button to import Discogs releases to MusicBrainz and add links to matching MusicBrainz entities for various Discogs entities (artist,release,master,label)
-// @version      2026.8.26.1
+// @version      2026.8.29
 // @namespace    http://userscripts.org/users/22504
 // @downloadURL  https://raw.githubusercontent.com/murdos/musicbrainz-userscripts/master/discogs_importer.user.js
 // @updateURL    https://raw.githubusercontent.com/murdos/musicbrainz-userscripts/master/discogs_importer.user.js
@@ -57,9 +57,6 @@ $(document).ready(function () {
     // it should be possible to react to pjax events
     $('div#pjax_container').attr('id', 'pjax_disabled');
 
-    // Display links of equivalent MusicBrainz entities
-    insertMBLinks(current_page_key);
-
     // Add an import button in a new section in sidebar, if we're on a release page
     let current_page_info = link_infos[current_page_key];
     if (current_page_info.type === 'release') {
@@ -99,6 +96,9 @@ $(document).ready(function () {
                 LOGGER.error('AJAX error thrown: ', errorThrown);
             },
         });
+    } else {
+        // Display links of equivalent MusicBrainz entities
+        insertMBLinks(current_page_key);
     }
 });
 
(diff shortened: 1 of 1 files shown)


jesus2099:
I don't know why this auto test is failing, as `eslint` showed me no errors, locally, on my only modified file (discogs_importer.user.js).

arsinclair:
It's failing because of the missing formatting. You can run `pnpm check` on your local and see the same thing that CI checks for.

I haven't looked at the code yet.

jesus2099:
I typed `npx pnpm check` and it seems to be related to another file than the one I have changed:

```
file:///C:/Users/m319937/AppData/Local/npm-cache/_npx/e6e0ed1aca658cae/node_modules/pnpm/dist/pnpm.mjs:159821
import { styleText } from "node:util";
         ^^^^^^^^^
SyntaxError: The requested module 'node:util' does not provide an export named 'styleText'
```

arsinclair:
> SyntaxError: The requested module 'node:util' does not provide an export named 'styleText

I'm honestly not sure about this error. It seems this might be happening if your NodeJS version is below 24.

Anyway, you can just run the formatter script and it should normalise everything.

arsinclair:
> I typed `npx pnpm check`

Also I don't know if this command is going to work, but if pnpm is installed on your system, there's no need to run it with `npx`. It is not an installable dependency of the current project, and it should be ran as a system-wide or user-wide binary.

arsinclair (commented):
I ran an AI code checker against the changes here and manually validated all the findings.

1. **all MB links disappear when the Discogs API fails**  
   `insertMBLinks()` now runs only from `insertMBSection()` after a successful API request (`discogs_importer.user.js:74, 99–101, 603–605`). A rate limit, network error, API outage, or parsing exception therefore prevents every MusicBrainz link from appearing on release pages—not just the release-group search link. Previously these links were independent of the import API request.
2. **arbitrary release titles aren’t URL/HTML encoded**  
   Lines 313–324 now pass the API title into `MBImport.searchUrlFor()`, which applies Lucene escaping but not `encodeURIComponent()`. For example, `Rock & Roll` produces:

   ```text
   ?query=Rock & Roll&type=release_group&indexed=1
   ```

   The browser interprets the query as `Rock ` plus an extra ` Roll` parameter. Double quotes in a title can also corrupt the interpolated HTML attribute.

A safer approach is to retain the original early `insertMBLinks()` call, then update only the sidebar search link’s `href` once the release title arrives. The URL should be constructed using encoded query parameters or assigned through the DOM rather than HTML interpolation.

jesus2099:
2. 

It may be a pre-existing bug because I don't think `&this.text()` returned some URI component encoded text, either

1.

I don't have much time, at the moment. Let's keep this aside for a while...

IMO, 1 is not that important, and the MB links should be better fixed anyway, as they regularly don't appear, because of some late Discogs DOM changes, already. Similar to how MBS React hydrate redraws DOM and wipes out all userscripts. 
My further delaying incidentally solved a portion of these occurrences. 😉

The best fix for 1, that I initially tried to add was: 
Add this RG search link only when the release (or even master) title is known, after the API call.

Because the RG search for _See all versions_ doesn't make sense and is of no use, even temporarily.

But the code is not mine and I don't get it totally (I'm not used to jQuery and such). I had not enough time to spend on this, and went to a simpler change that fixed it for me. I didn't think those MB links could still be useful when the import is failing (API down). But I guess you're right.

jesus2099:
I'm marking this as draft. I'll keep this version for my own usage, until I have time to write a better version for this repo. Or until someone else does it, I don't mind.

arsinclair:
If you don't have time, I can probably fix them some time before the end of the year. The requested changes are not that big of a deal, but also I don't work with the discogs script that much and its logic is already too convoluted, so it will take me a while to wrap my head around it.

Regarding problem #2 it seems it is a tiny fix, we already have a function that correctly escapes unescaped characters for lucene.

Regarding #1: if MB API is responding with errors (which happens every day now btw) it still may be helpful if the script renders the icons that are cached in the local storage, but with this change they would disappear.

> the MB links should be better fixed anyway, as they regularly don't appear, because of some late Discogs DOM changes, already. Similar to how MBS React hydrate redraws DOM and wipes out all userscripts.

I agree, I saw that bug and never really got time to pin it down. It should definitely be fixed too, as it's very annoying.

The author tag is not filled here, but from the commit history alone I can see that @murdos is the author of the script, so he might be better equipped reviewing this and evaluating whether the problems outlined are actually important enough, or can be fixed later on their own.

murdos:
I wrote this script a long time ago (more than 15 years), and I don't use it since more that 10 years, so I really don't have any idea.
Maybe @zas has one, since he maintained it at some point.

PR #1070, opened 2026-09-05
fix(deezer_importer): Deezer importer script requires a page refresh to work (#699)

### LINKED ISSUE #699: Deezer importer script requires a page refresh to work
Similar to https://github.com/murdos/musicbrainz-userscripts/issues/698, a full page refresh is required in order for the script to start. We need to fix that.

chaban-mb:
_Note: Code and description were written with AI-tools_

### Description
Fixes #699. Supports client-side SPA navigation on Deezer so that the importer automatically loads on album pages and removes previously mounted buttons when navigating away, without requiring a manual page refresh.

### Changes & Rationale
- **Broaden match pattern (`meta.json`)**: Updated `@match` to `https://www.deezer.com/*` so the userscript activates across all Deezer routes. When users navigate client-side from home/search/artist pages to an album, the script is already running to handle the transition.
- **Listen for SPA navigation (`index.ts`)**: Extracted page processing into `processPage()` and subscribed to route changes using `subscribeToSPANavigation` so album detection runs on every client-side transition.
- **Teardown previously mounted UI (`cleanup()`)**: Track inserted elements in `mountedElements` and remove them upon navigation. Because Deezer reuses the same toolbar DOM node across navigations, this prevents button duplication and ensures buttons are removed when navigating to non-album pages.
- **Race condition guard (`runId`)**: Pass a monotonic `runId` to `waitForEl` and the API response handler to safely discard late DOM elements or out-of-order API responses if a user rapidly navigates between albums.

### Verification
- `pnpm type-check` (`tsgo`): 0 errors
- `pnpm lint` (`oxlint --type-aware` & `eslint .`): 0 errors, 0 warnings
- `pnpm build`: successfully compiled
- `pnpm test`: 62 tests passed
- **Live Verification**: Tested on live `deezer.com` in both desktop (`[data-testid="toolbar"]`) and mobile viewports (`[data-tracking-label="main-CTA"]`), confirming clean teardown on navigation, no button duplication, and automatic mounting across client-side album transitions.

### DIFF
--- a/src/userscripts/deezer_importer/index.ts
+++ b/src/userscripts/deezer_importer/index.ts
@@ -1,23 +1,43 @@
 import { Logger, LogLevel } from '~/lib/logger';
 import { MBImport } from '~/lib/mbimport';
 import { MBImportStyle } from '~/lib/mbimportstyle';
+import { subscribeToSPANavigation } from '~/lib/shared/spa-navigation';
 
 import { getDeezerReleaseData } from './utils/getDeezerReleaseData';
 import { parseDeezerRelease } from './utils/parseDeezerRelease';
 
 const LOGGER = new Logger('deezer_importer', LogLevel.INFO);
 
-function waitForEl(selector: string, callback: () => void): void {
-    if (document.querySelector(selector)) {
-        callback();
+let currentRunId = 0;
+let mountedElements: HTMLElement[] = [];
+
+function cleanup(): void {
+    mountedElements.forEach(el => {
+        el.remove();
+    });
+    mountedElements = [];
+}
+
+function waitForEl(selector: string, runId: number, callback: (el: HTMLElement) => void): void {
+    if (runId !== currentRunId) {
+        return;
+    }
+    const el = document.querySelector<HTMLElement>(selector);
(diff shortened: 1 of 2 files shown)


src/userscripts/deezer_importer/index.ts
```diff
@@ -1,23 +1,43 @@
...
+let mountedElements: HTMLElement[] = [];
+
+function cleanup(): void {
+    mountedElements.forEach(el => {
+        el.remove();
+    });
+    mountedElements = [];
+}
+
+function waitForEl(selector: string, runId: number, callback: (el: HTMLElement) => void): void {
```
arsinclair, line 21: `callback: (el: HTMLElement)`

nice, I had this optimisation in mind for quite some time!

arsinclair (approved):

arsinclair:
Everything looks good.

src/userscripts/deezer_importer/index.ts
chaban-mb: It wasn't my idea. Thank Gemini :P

